PR #11, opened 2019-02-27
[ObjectsTable] Fix pagination next/prev

### LINKED ISSUE #10: ObjectsTable improvements
From https://github.com/EyeSeeTea/metadata-synchronization/pull/21

- [x] Fix pagination -> #11 

High Priority
- [x] ObjectsTable: Hide checkboxes if there are no fields with multiple set to true.
- [x] Apply branch fix/calculate-page-value
- [x] ObjectsTable: We have `initialSorting` as a required prop, make it optional and assume first sortable column ASC is the default order.
- [x] Update readme for ObjectsTable with explanation regarding prop 'key' to update it.

Low Priority
- [x] Reduce marginTop by 10px, too much space.
- [x] Change model prop to optional (See https://github.com/EyeSeeTea/metadata-synchronization/issues/7 -> Remove model prop from instance-configurator)
- [x] Hide details panel if table refreshes otherwise it can be confusing for the user.

tokland:
Fixes #10 -> Fix pagination

The old code was expecting a D2 pager (with `hasNextPage`/`hasPreviousPage`). Now we simply calculate the values from the current page / max page.

(Not yet published, will do after PRs approved & merged)

### DIFF
--- a/src/objects-table/ObjectsTable.js
+++ b/src/objects-table/ObjectsTable.js
@@ -199,12 +199,16 @@ class ObjectsTable extends React.Component {
     };
 
     getPaginationProps() {
-        const currentlyShown = calculatePageValue(this.state.pager, this.props.pageSize);
         const { pager } = this.state;
 
+        const isValidPage = _(["page", "pageCount", "total"])
+            .difference(_.keys(pager || {}))
+            .isEmpty();
+        if (!isValidPage) return {};
+
         return {
-            hasNextPage: () => Boolean(pager.hasNextPage) && pager.hasNextPage(),
-            hasPreviousPage: () => Boolean(pager.hasPreviousPage) && pager.hasPreviousPage(),
+            hasNextPage: () => pager.page < pager.pageCount,
+            hasPreviousPage: () => pager.page > 1,
             onNextPageClick: () => {
                 this.setState(
                     {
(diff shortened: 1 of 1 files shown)


SferaDev (approved):

SferaDev:
Looks good to me, to simplify it even more we could also calculate (and not require the field) the pageCount as it's always Math.ceil(total / pageSize). What do you think @tokland ?

src/objects-table/ObjectsTable.js
```diff
@@ -199,12 +199,17 @@ class ObjectsTable extends React.Component {
...
 
     getPaginationProps() {
-        const currentlyShown = calculatePageValue(this.state.pager, this.props.pageSize);
         const { pager } = this.state;
 
+        const isValidPage = _(["page", "total"])
+            .difference(_.keys(pager || {}))
+            .isEmpty();
+        if (!isValidPage) return {};
+        const pageCount = Math.ceil(pager.total / pager.pageSize);
```
SferaDev, line 208: It should be this.props.pageSize, right?

tokland: Right! It was working for me because I do have pageSize also on the pager, but should get from props. I'll push to dev.

tokland:
d2-ui-components 0.0.9 published. Note that there is a delay on the webpage (it will show 0.0.8 for some minutes), but it should already be on the npm registry ready to be used.

PR #16, opened 2019-03-11
ObjectsTable: Add onSelectionChange

### LINKED ISSUE #14: Allow app to receive selection changes
Add onSelectionChange prop to handle state

SferaDev:
### :pushpin: References

* **Issue:** Closes #14 

### 💻 Status

- Ready to merge

- Tested with metadata-sync

### :memo: Implementation

- Selection is still in a controlled state

- Objects Table notifies App with new selection with onSelectionChange prop

- Selection is deep cloned so the internal reference is lost outside ObjectsTable

### 🎨 Screenshots

image

### 📓 Test example

**Metadata Synchronization**

```js
--- a/src/components/sync-configurator/BaseSyncConfigurator.jsx
+++ b/src/components/sync-configurator/BaseSyncConfigurator.jsx
@@ -40,6 +40,7 @@ export default class BaseSyncConfigurator extends React.Component {
                     initialSorting={model.getInitialSorting()}
                     actions={this.actions}
                     list={list}
+                    onSelectionChange={(selection) => console.log({selection})}
                 />
             </div>
```

### DIFF
--- a/src/objects-table/ObjectsTable.js
+++ b/src/objects-table/ObjectsTable.js
@@ -129,7 +129,7 @@ class ObjectsTable extends React.Component {
             sorting: this.props.initialSorting,
             searchValue: null,
             detailsObject: null,
-            selection: [],
+            selection: new Set(),
         };
     }
 
@@ -142,7 +142,7 @@ class ObjectsTable extends React.Component {
     }
 
     notifySelectionChange = () => {
-        this.props.onSelectionChange(_.clone(this.state.selection));
+        this.props.onSelectionChange(Array.from(this.state.selection));
     };
 
     async getObjects({ clearPage = true } = {}) {
@@ -153,13 +153,16 @@ class ObjectsTable extends React.Component {
         const pagination = { page: newPage, pageSize: pageSize, sorting };
         const { pager, objects } = await list(d2, filters, pagination);
 
-        this.setState({
-            isLoading: false,
-            pager: pager,
-            dataRows: objects,
-            page: newPage,
-            selection: [],
-        }, this.notifySelectionChange);
+        this.setState(
+            {
+                isLoading: false,
+                pager: pager,
+                dataRows: objects,
+                page: newPage,
+                selection: new Set(),
+            },
+            this.notifySelectionChange
+        );
     }
 
     onSearchChange = value => {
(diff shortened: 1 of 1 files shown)


SferaDev:
Topics to be discussed:

Page reloading:

- User selects A
- User reorders columns
- A is not selected anymore

src/objects-table/ObjectsTable.js
```diff
@@ -260,14 +265,12 @@ class ObjectsTable extends React.Component {
...
+        const selectedHeaderChecked =
+            !_.isEmpty(dataRows) && dataRows.every(row => selection.includes(row.id));
 
         const allColumns = this.getColumns(columns, selectedHeaderChecked);
 
-        const activeRows = _(rows)
-            .keyBy("id")
-            .at(dataRows.filter(dr => dr.selected).map(dr => dr.id))
-            .value();
+        const activeRows = selection.map(id => rows.find(dr => dr.id === id));
```
tokland: map -> find is O(n^2). To avoid this I usually resort to a keyBy + at, which is O(n)

tokland: Also, about this.state.selection. Does it make sense to use a Set instead of an array? Check inclusions (`myset.has()`) are O(1) instead of O(n) this.

src/objects-table/ObjectsTable.js
```diff
@@ -177,20 +180,22 @@ class ObjectsTable extends React.Component {
         ev.stopPropagation();
 
         const { selection } = this.state;
-        const deletedIds = _.remove(selection, id => id === obj.id);
-        if (deletedIds.length === 0) selection.push(obj.id);
+        const found = selection.delete(obj.id);
+        if (!found) selection.add(obj.id);
```
tokland, line 184: I think JS sets operations are mutable, right? We need immutable updates to play nice with React comparisons. Maybe Array.from + append or remove + new Set()?

tokland: Not sure about this logic of deleting followed by adding. Anyway, I guess it will be be refactored when doing it immutable.

src/objects-table/ObjectsTable.js
```diff
@@ -266,11 +271,14 @@ class ObjectsTable extends React.Component {
...
-            !_.isEmpty(dataRows) && dataRows.every(row => selection.includes(row.id));
+            !_.isEmpty(dataRows) && dataRows.every(row => selection.has(row.id));
 
         const allColumns = this.getColumns(columns, selectedHeaderChecked);
 
-        const activeRows = selection.map(id => rows.find(dr => dr.id === id));
+        const activeRows = _(rows)
+            .keyBy("id")
+            .at(dataRows.filter(dr => selection.has(dr.id)).map(dr => dr.id))
+            .value();
```
tokland, line 281: Cool!

src/objects-table/ObjectsTable.js
SferaDev: The logic is a toggle:
- Try to delete from Set, it returns if it existed
- Create in Set if it didn't exist before attempting to delete

I guess creating a shallow copy of the Set would suffice to avoid mutating the state.
```js
const selection = new Set(this.state.selection);
```

tokland (approved):

PR #20, opened 2019-03-26
Add loading mask to d2-ui-components

SferaDev:
### :memo: Implementation

- Add a Loading Mask Provider

### :art: Screenshots

image


README.md
```diff
@@ -121,6 +121,44 @@ MyComponent.propTypes = {
...
+To use it, create a HOC with `withLoading`, add `loading` to your propTypes, and show the mask using the functions `props.loading.show()`.
+
+```
+import { withLoading } from "d2-ui-components";
+import PropTypes from "prop-types";
+
+const MyComponent = ({name, loading}) => (
+    <div>
+        <a onClick={() => loading.show()}>Show loading mask</a>
+        <a onClick={() => loading.show(false)}>Hide loading mask</a>
```
tokland, line 166: `loading.show(boolean)` is very useful, but can we have also `loading.hide()` for the simple case?

src/loading/LoadingConsumer.js
```diff
@@ -0,0 +1,81 @@
...
+    },
+    divider: {
+        margin: "40px"
+    }
+});
+
+const LoadingConsumer = props => {
+    const { classes } = props;
+
+    const showMessage = () => {
```
tokland: Unused function.

src/loading/LoadingConsumer.js
```diff
@@ -0,0 +1,81 @@
...
+                            />
+                            <Divider
+                                className={classes.divider}
+                                variant="middle"
+                                hidden={!message || message.trim().length === 0}
+                            />
+                            <Typography
+                                className={classes.message}
+                                variant="h6"
+                                hidden={!message || message.trim().length === 0}
```
tokland: Used two times, maybe move it a to local variable `hidden`?

src/loading/LoadingConsumer.js
```diff
@@ -0,0 +1,81 @@
+import React from "react";
```
tokland: Should be a jsx file. I see why you have a js file, I had it wrong for the snack: `src/snackbar/SnackbarConsumer.js`, could be changed also there.

src/loading/LoadingConsumer.js
```diff
@@ -0,0 +1,81 @@
...
+import PropTypes from "prop-types";
+
+import { withStyles } from "@material-ui/core/styles";
+import LoadingContext from "./context";
+import CircularProgress from "@material-ui/core/CircularProgress";
+import Typography from "@material-ui/core/Typography";
+import Divider from "@material-ui/core/Divider";
+
+const styles = theme => ({
```
tokland: eslint does not like non-used arguments, `_theme` should do the trick.

src/loading/LoadingProvider.js
```diff
@@ -0,0 +1,55 @@
+import React, { Component } from "react";
```
tokland: Same, jsx file.

src/loading/LoadingProvider.js
```diff
@@ -0,0 +1,55 @@
...
+import LoadingContext from "./context";
+import LoadingConsumer from "./LoadingConsumer";
+
+export default class SnackbarProvider extends Component {
```
tokland: LoadingProvider

tokland (commented):
The code (and the snapshot, nice!) looks good to me, just some minor things.

Btw, Is there a branch in metadata-sync that uses it, so we can check how it works? Sometime we should create a simple dirty app with all the components on one page so we can them easily.

README.md
SferaDev: Sounds good, will add.

src/loading/LoadingConsumer.js
SferaDev: Yeah I just started the branch off Snackbar with shallow copy. Will change them in this PR now

src/loading/LoadingConsumer.js
SferaDev: Okay I'll move to render variable

tokland: Now that we are at it, isn't equivalent to: `!message || !message.trim()`?

SferaDev:
> Btw, Is there a branch in metadata-sync that uses it, so we can check how it works? Sometime we should create a simple dirty app with all the components on one page so we can them easily.

Not yet but we should have it before Friday. Anyway I think the best approach is creating a [StoryBook](https://github.com/storybooks/storybook). It's really cool and now it's like the state-of-the-art solution for ReactJS examples.

tokland:
> Anyway I think the best approach is creating a [StoryBook](https://github.com/storybooks/storybook). 

Wow, it really looks nice. I think it still makes sense to  have an `example` CRA in the repo itself, but it looks good. Maybe you can show us on the next team meeting.

tokland (approved):
I have not tested functionally, I'll wait for CRA example (or metadata-sync test branch), but the code looks very good to me.

adrianq (approved):
I tested all the features in the multiple metadata-sync PR. Awesome work guys!

PR #21, opened 2019-03-27
Create datePickerFilter to use with ObjectsTable

matiasgarcia91:
This is a DatePicker formatted to look good as an ObjectsTable filter. `label` is removed in favor of `placeholder` and styles differ from the DatePicker form field

image
image

### DIFF
--- a/src/date-table-filter/DateTableFilter.jsx
+++ b/src/date-table-filter/DateTableFilter.jsx
@@ -1,0 +1,92 @@
+import React, { PureComponent } from "react";
+import PropTypes from "prop-types";
+import moment from "moment";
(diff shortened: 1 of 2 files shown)


matiasgarcia91:
suggestions? it's pretty similar to the other DatePicker but it overrides some different MUI styles and uses a placeholder instead of label. Since Adrian commented we will be using it practically in every table I thought it was better to have it on it's own but maybe it's too similar

matiasgarcia91:
I'll update the README if you guys think this component is worth including

tokland:
I'd vote for keeping just one component. If necessary we can add some props to DatePicket so it does what we need from a form or working as a filter field.

Issue #8, opened 2019-02-26
Components to abstract


tokland:
High Priority
- [x] [Migrate] Confirmation dialog (metadata-sync has the latest changes). See https://github.com/EyeSeeTea/vaccination-app/blob/development/src/components/confirmation-dialog/ConfirmationDialog.jsx
   - [x] Add optional props for button labels
   - [x] Integrate new confirmation dialog into dialog handler. Substitute Dialog component in render
   - [x] Rename components:
      Dialog Handler -> DialogButton
      Confirmation Dialog -> ConfirmationDialog

Low Priority
- [x] [Migrate] Wizard (from vaccination-app)
- [ ] [Improvement] expose DataTable: (clean up props -specially actions, should be as simple as ObjectsTable-, expose it in index.ts, and add docs to README).

adrianq:
Not relevant anymore. We are currently following a different approach in most of the projects

PR #22, opened 2019-03-27
Abstract Confirmation dialog and update Dialog Button

SferaDev:
### :memo: Implementation

- Abstract Confirmation Dialog from metadata-sync

- Improve the code for generic usage

- Update Dialog button to use Confirmation Dialog as discussed on #8

### DIFF
--- a/src/confirmation-dialog/ConfirmationDialog.jsx
+++ b/src/confirmation-dialog/ConfirmationDialog.jsx
@@ -22,32 +22,39 @@ class ConfirmationDialog extends React.Component {
     };
 
     render() {
-        const { title, description, onSave, onCancel, isOpen, children, cancelText, saveText, ...other } = this.props;
+        const {
+            title,
+            description,
+            onSave,
+            onCancel,
+            isOpen,
+            children,
+            cancelText,
+            saveText,
+            ...other
+        } = this.props;
 
         return (
             <React.Fragment>
-                <Dialog
-                    open={isOpen}
-                    onClose={onCancel ? onCancel : () => null}
-                    {...other}
-                >
-                    <DialogTitle>
-                        {title}
-                    </DialogTitle>
+                <Dialog open={isOpen} onClose={onCancel ? onCancel : () => null} {...other}>
+                    <DialogTitle>{title}</DialogTitle>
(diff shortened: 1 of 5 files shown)


src/confirmation-dialog/ConfirmationDialog.jsx
```diff
@@ -0,0 +1,65 @@
...
+            ...other
+        } = this.props;
+
+        return (
+            <React.Fragment>
+                <Dialog open={isOpen} onClose={onCancel} {...other}>
+                    <DialogTitle>{title}</DialogTitle>
+
+                    <DialogContent>
+                        {description && { description }}
```
tokland: I get an error on this: `Objects are not valid as a React child (found: object with keys {description})`. Maybe `description ? description : null`, or even `description || null`. (`null` is important, as we know, React does not accept `undefined` as node)

src/confirmation-dialog/ConfirmationDialog.jsx
```diff
@@ -0,0 +1,65 @@
...
+            onCancel,
+            isOpen,
+            children,
+            cancelText,
+            saveText,
+            ...other
+        } = this.props;
+
+        return (
+            <React.Fragment>
```
tokland: Can we remove the React.Fragment, as we only have one child?

src/index.ts
```diff
@@ -17,7 +18,8 @@ export {
     MultiSelector,
     SimpleCheckBox,
     SearchBox,
-    DialogHandler,
+    DialogButton,
+    ConfirmationDialog,
```
tokland, line 22: Since we are exposing ConfirmationDialog, should we an entry in the README?

tokland (commented):
I added a rule for eslint, for _ignoreThisVar to work. Also some comments:

src/confirmation-dialog/ConfirmationDialog.jsx
SferaDev: I finally added something like this:

```js
{description && <DialogContentText>{description}</DialogContentText>}
```

tokland (approved):
All good to me!

PR #23, opened 2019-03-27
MultiSelector: Allow selection to be uncontrolled

SferaDev:
### :memo: Implementation

- Make ordered optional and default to true

- Add a uncontrolled state selection

- If selection is overrided from parent through props ignore enter controlled state

### DIFF
--- a/src/multi-selector/MultiSelector.jsx
+++ b/src/multi-selector/MultiSelector.jsx
@@ -20,8 +20,10 @@ const optionsPropType = PropTypes.arrayOf(
 );
 
 class MultiSelector extends React.Component {
+    isControlled = !!this.props.selected;
+
     state = {
-        selected: []
+        selected: [],
     };
 
     static propTypes = {
@@ -36,6 +38,7 @@ class MultiSelector extends React.Component {
     static defaultProps = {
         height: 300,
         ordered: true,
+        selected: undefined,
     };
 
     // Required by <GroupEditor>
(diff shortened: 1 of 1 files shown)


tokland (commented):
I tested with the component in vaccionation (controlled), there is a problem with the initial selected.

I've committed a proposal to keep the controlled/uncontrolled mode more declarative. Using the state.selected on a controlled component may lead to problems (two sources of truth). 

Let me know if it looks good for you.

src/multi-selector/MultiSelector.jsx
```diff
@@ -49,28 +52,44 @@ class MultiSelector extends React.Component {
...
+
+        if (this.isControlled) {
+            this.props.onChange(selectedUpdated);
+        } else {
+            this.setState({ selected: selectedUpdated });
+            this.props.onChange(selectedUpdated);
+        }
+
+        // <GroupEditor> requires the props assign/unassign/orderChange to return a promise
+        return Promise.resolve();
```
SferaDev, line 66: Adding async to the method already returns a promise implicitly and makes the code simpler

src/multi-selector/MultiSelector.jsx
```diff
@@ -49,28 +52,44 @@ class MultiSelector extends React.Component {
...
-        this.props.onChange(selected);
+    getSelected = () => {
+        return this.isControlled ? this.props.selected : this.state.selected;
+    };
+
+    updateSelected = updateFn => {
+        const selected = this.getSelected();
+        const selectedUpdated = updateFn(selected);
+
+        if (this.isControlled) {
```
SferaDev: Can be simplified to always setState and onChange. State will be ignored when controlled mode

src/multi-selector/MultiSelector.jsx
tokland: I'd prefer to be explicit. I had written this code only one month ago and had already forgotten about this, so today I removed the `async` on my refactor, because apparently it's not needed, and then you get a very cryptic error. Then I remembered why we had the async :)

SferaDev: Ah, okay. No worries then, re-adding it!

src/multi-selector/MultiSelector.jsx
tokland: I considered it...  but now that I think of it, a change of state in React means a re-render. When in fact this.state.selected won't be used in render because it's controlled. So it's a bit weird. I'd leave it.

SferaDev: Okay

SferaDev:
Looks good to me, tested with https://github.com/EyeSeeTea/metadata-synchronization/pull/49 and everything working as expected.

tokland (approved):

PR #24, opened 2019-03-28
Extend datePicker with isFilter prop

matiasgarcia91:
Extended DatePicker to receive `isFilter` prop and style itself accordingly


tokland (approved):
Added some type restriction on color (3bbc6341). The prop `isFilter` is a bit restrictive, but if this is our custom use (as a field form or as a filter), it's convenient.

SferaDev (approved):

adrianq:
@matiasgarcia91 you are still working on the feature in metadata-sync that uses this new prop, right?

matiasgarcia91:
Yes, but the date-picker part is finished so this is good to go

PR #33, opened 2019-04-19
Fix multiple objects callback handler

### LINKED ISSUE #28: ObjectsTable selection lost with clicks
When clicking outside the table or in other rows the whole selection gets lost.

tokland:
### :pushpin: References

* **Issue:** Closes #28 

### :memo: Implementation

Multiple selection had a bug. Refactored to use always an array of objects and only at the final callback send a single object if `multiple` option is disabled.


SferaDev:
Tested locally and everything seems to work fine.

SferaDev (approved):

PR #34, opened 2019-04-24
Objects Table Selection improvements

### LINKED ISSUE #12: Improve Object Selection
When a re-render happens ObjectsTable forgets about previously selected items.

Proposal: Move selection id array out of ObjectsTable and make it a required prop. That way it will be block-captured and the selection will persist across re-renders.

This proposal has the side-benefit of being able to read the currently selected items outside Objects Table and perform other operations as metadata-sync requires.

Added this issue outside the Improvements card because of its severity.

SferaDev:
### :pushpin: References

* **Issue:** Closes #12 

### :memo: Implementation

- Streamline Table Header Checkbox with d2-ui-components SimpleCheckbox instead of mui.

- Get all objects ids on page reload and save them to state.

- When changing pages do not forget previous selection.

- Add a notification messages zone as caption of the objects table.

- Rename selection methods to represent the action they accomplish.

- If selection in all pages is implemented by the list method and all items in current page are selected show a message to select all the ids in all pages. (Note that if already selected the message is not shown).

- If there are selected items in other pages, not visible to the user show a message to clear the selection in all pages.

- In some scenarios both messages can appear and are grouped in the same notification panel.

### 🎨 Screenshots

image

image

image

image

image

### :bookmark_tabs: Others

- Test App: [metadata-synchronization.zip](https://github.com/EyeSeeTea/d2-ui-components/files/3118693/metadata-synchronization.zip)


i18n/en.pot
```diff
@@ -50,6 +50,29 @@ msgstr ""
 msgid "selected"
 msgstr ""
 
+msgid "All {{count}} items on this page are selected."
+msgid_plural "All {{count}} items on this page are selected."
+msgstr[0] ""
+msgstr[1] ""
```
tokland: So this is how plurals work, good to know! 

To sum it up: When a `i18n.t(msg, {count: N})` is found (`count` is special key), gettext creates a msgid and msgid_plural, and -in the general case, see `Plural-Forms`-, `msgstr[0]` is the singular text and `msgstr[1]` the plural text.

https://www.gnu.org/software/gettext/manual/html_node/Translating-plural-forms.html

i18n/en.pot
```diff
@@ -50,6 +50,29 @@ msgstr ""
...
+
+msgid "Clear selection"
+msgstr ""
+
+msgid "There are {{count}} items selected on other pages."
+msgid_plural "There are {{count}} items selected on other pages."
+msgstr[0] ""
+msgstr[1] ""
```
tokland: We are not creating custom messages for singular/plural, is that deliberate?

The fr/es translations for those keys seem to be missing, after a `yarn update-po` they appear.

src/objects-table/ObjectsTable.js
```diff
@@ -99,9 +98,9 @@ class ObjectsTable extends React.Component {
     getColumns(columns, selectedHeaderChecked) {
         const { model } = this.props;
         const selectedColumnContents = (
-            <Checkbox
+            <SimpleCheckBox
```
tokland, line 101: Yep, it makes sense to use the same checkbox everywhere.

src/objects-table/ObjectsTable.js
```diff
@@ -248,15 +268,63 @@ class ObjectsTable extends React.Component {
...
+            });
+        } else if (selectionInOtherPages.length > 0) {
+            messages.push({
+                message: i18n.t("There are {{count}} items selected on other pages.", {
+                    count: selectionInOtherPages.length,
+                }),
+                link: i18n.t("Clear selection"),
+                action: this.clearSelection,
+            });
+        }
```
tokland: It's ok as it is, no worries :) But just as a note, to make it functional, you may write:

```
const messages = _.compact([
  condition && message, // or condition ? message : null
  condition2 && message2,
  ...
])
```

Granted, it may not look so nice.

src/objects-table/ObjectsTable.js
```diff
@@ -130,6 +129,7 @@ class ObjectsTable extends React.Component {
             searchValue: null,
             detailsObject: null,
             selection: new Set(),
```
tokland, line 131: Selection is cleared if a row is clicked and the detail box is open, is that intentional? Also if the contextual menu for a row is open.

src/objects-table/ObjectsTable.js
```diff
@@ -248,15 +268,63 @@ class ObjectsTable extends React.Component {
...
+        }
+
+        const selectionInOtherPages = _.difference(
+            Array.from(selection),
+            dataRows.map(dr => dr.id)
+        );
+
+        if (selectedAllInAllPages) {
+            messages.push({
+                message: i18n.t("There are {{count}} items selected in all pages.", {
```
tokland: There is an edge case with a un-intuitive. message  Let's imagine I have all items (2010) selected, then I see: `There are 2010 items selected in all pages`. I unselect one, now I see `There are 1990 items selected on other pages`. That's correct, but maybe we could show also the total? `There are 2009 items selected (1990 in other pages)`. Not sure if this makes sense, just as an idea.

tokland (commented):
That's a nice feature to have, cool! I didn't found any problem on the selection, it worked as expected. So just some comments:

src/objects-table/ObjectsTable.js
SferaDev: No, that's a bug!! Thanks :)

src/objects-table/ObjectsTable.js
SferaDev: Makes sense

SferaDev:
- Updated the code to allow plurals where needed (only one phrase actually) and removed the plural pot detection on the rest.

- Message building is now purely functional and has be refactored.

- Conditions are now more robust, previously when filtering we could potentially see an invalid select all pages message.

PR #37, opened 2019-05-13
Add props to Object table to permit a label in the fab button

matiasgarcia91:
image

### DIFF
--- a/src/objects-table/ObjectsTable.js
+++ b/src/objects-table/ObjectsTable.js
@@ -66,12 +66,12 @@ class ObjectsTable extends React.Component {
         customFiltersComponent: PropTypes.func,
         customFilters: PropTypes.object,
         onSelectionChange: PropTypes.func,
-        createButtonLabel: PropTypes.node,
+        buttonLabel: PropTypes.node,
     };
 
     static defaultProps = {
         onSelectionChange: () => {},
-        createButtonLabel: null,
+        buttonLabel: null,
     };
 
     constructor(props) {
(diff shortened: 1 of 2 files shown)


SferaDev:
If we follow the approach from Metadata Sync, we decided to customize the button in the app instead of the ui library.

tokland (approved):
Tested with vaccination-app, works as expected.

tokland:
Prop `createButtonLabel` changed to `PropTypes.node`, so it can be anything renderable (typically: string or React element):

https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes

@adrianq, @SferaDev  Ready for review. Once merged, I'll update to npmjs so we can use it in metadata-sync.

SferaDev (approved):
Tested on metadata-sync

image

image

src/objects-table/ObjectsTable.js
```diff
@@ -421,7 +430,7 @@ class ObjectsTable extends React.Component {
                     ) : null}
                 </div>
 
-                {onCreate && <ListActionBar onClick={onCreate} />}
+                {onCreate && <ListActionBar onClick={onCreate} label={createButtonLabel} />}
```
SferaDev: *Minor:* I know it's a breaking change (should be avoided if third parties used the library) but wouldn't it make sense something like:

```
{floatingButtonAction && <ListActionBar onClick={floatingButtonAction} label={floatingButtonLabel} />}
```

I understand it originally meant "Create" but now it's generic enough to be called something else.

SferaDev (approved):

adrianq:
@tokland Works fine. I have just changed the prop to buttonLabel as suggested by @SferaDev. I have already taken care of propagating this change to vaccination repo. @SferaDev can you do the same with metadata sync? @tokland Whenever you have a change if you can release to npm...

SferaDev:
@adrianq Updated accordingly https://github.com/EyeSeeTea/metadata-synchronization/pull/92

PR #39, opened 2019-05-28
Allow hiding default search box from ObjectsTable

SferaDev:
Allow hiding the search box when the page does not need a string search. Right now we need it for Notifications Page on metadata-sync where we will only host a CustomFilter with a dropdown.

image


SferaDev:
LGTM

tokland:
As discussed with adrianq, used this PR to change the props. 

vaccination-app: https://github.com/EyeSeeTea/vaccination-app/pull/157

PR #41, opened 2019-06-11
Import Wizard from vaccination-app

SferaDev:


tokland (commented):
It works without problems. I guess we should add it to the README.

tokland:
Added a couple of commits, one about i18n, the other about the problem @SferaDev reported. Finally, it was a bug in nano-memoize v1.1.3:

https://github.com/anywhichway/nano-memoize/issues/7

PR #42, opened 2019-06-18
Add initial selection to Objects Table

SferaDev:
This allows to pre-load the Objects Table with selected items.

Use case: We now have an ObjectsTable in a Wizard, on edit (or even pagination) we load the selected items from the stored state.

Example (the elements were selected programmatically by the stored state):

image

### DIFF
--- a/src/objects-table/ObjectsTable.js
+++ b/src/objects-table/ObjectsTable.js
@@ -65,6 +65,7 @@ class ObjectsTable extends React.Component {
         customFiltersComponent: PropTypes.func,
         customFilters: PropTypes.object,
         onSelectionChange: PropTypes.func,
+        initialSelection: PropTypes.array,
         buttonLabel: PropTypes.node,
         hideSearchBox: PropTypes.bool,
     };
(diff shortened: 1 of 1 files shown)


src/objects-table/ObjectsTable.js
```diff
@@ -131,7 +133,7 @@ class ObjectsTable extends React.Component {
             sorting: this.props.initialSorting,
             searchValue: null,
             detailsObject: null,
-            selection: new Set(),
+            selection: new Set(this.props.initialSelection),
```
tokland, line 136: Let's say I pass an ID in `initialSelection` which is not in fact in allObjects, the messages about the selection won't be accurate. Is this a problem?

tokland (commented):
Cool! Just a comment/question.

src/objects-table/ObjectsTable.js
SferaDev: Not really, as we might have them hidden by the filters but that are in fact selected

tokland (approved):

PR #43, opened 2019-06-25
Set a default autoHide only for success messages

tokland:
On success messages it's ok to auto-hide the snackbar messages, but in any other level, the user needs time to read or maybe copy/paste its contents. Now the user will have to dismiss it on non-success messages.


SferaDev (approved):
Works fine on metadata-sync. Tested both scenarios:

- Info message that defaults to not hiding
- Info message with prop override that forces auto-hiding

PR #44, opened 2019-06-27
Add vertical align to data table

SferaDev:
Vertical alignment is not ideal when working with multiple lines, instead of fixing the position to absolute, force it to the middle of the row.

Before:

image

After:

image


tokland (approved):
It's apparent that the columns width is modified when there are multi-line fields, but it does not seem to be a problem.

SferaDev:
> It's apparent that the columns width is modified when there are multi-line fields, but it does not seem to be a problem.

Yeah, actually here we are overriding the getValue and adding an HTML unordered list with list items

PR #47, opened 2019-07-03
Add property isPrimary to actions

tokland:
Required by https://github.com/EyeSeeTea/vaccination-app/pull/222


adrianq:
Works perfectly! Keep in mind we need to release tomorrow and update the dependency in vaccination-app

PR #48, opened 2019-07-11
[OrgUnitsSelector] Add props listParams and rootIds

tokland:
Required by https://github.com/EyeSeeTea/vaccination-app/pull/237

### :memo: Implementation

This PR adds 2 props to the component `OrgUnitsSelector`:

- `listParams`. This value will be added to `d2.models.organisationUnits.list`, so we can use specific arguments supported by dhis ([docs](https://docs.dhis2.org/2.30/en/developer/html/dhis2_developer_manual_full.html#webapi_list_of_organisation_units)).

- `rootIds`. By default, all orgUnits were being displayed, now we can select a specific group. Note that we have an option in the endpoint `withinUserHierarchy` but this only allows to filter within the "data capture" orgUnits. When filtering by name, we do this filtering by JS (not possible to do both text/root inclusson filtering with a single API call).

### DIFF
--- a/src/org-units-selector/OrgUnitsSelector.jsx
+++ b/src/org-units-selector/OrgUnitsSelector.jsx
@@ -101,10 +101,18 @@ export default class OrgUnitsSelector extends React.Component {
         } else if (filter && !rootIds) {
             options = { filter, ...pagingOptions };
         } else if (filter && rootIds) {
+            // We cannot both filter by name and check inclusion on rootIds on the same request, so
+            // let's make a request filtering only by name and later check the rootIds
+            // in the response. Also, limit pageSize to avoid an uncontrolled big request.
             options = {
                 filter,
-                ...pagingOptions,
-                postFilter: orgUnit => rootIds.some(ouId => orgUnit.path.includes(ouId)),
+                paging: true,
+                pageSize: 1000,
+                postFilter: orgUnits =>
+                    _(orgUnits)
+                        .filter(orgUnit => rootIds.some(ouId => orgUnit.path.includes(ouId)))
+                        .take(pagingOptions.pageSize)
+                        .value(),
             };
         }
 
@@ -117,7 +125,8 @@ export default class OrgUnitsSelector extends React.Component {
 
         return d2.models.organisationUnits
             .list(listOptions)
-            .then(collection => collection.toArray().filter(options.postFilter || _.identity));
+            .then(collection => collection.toArray())
+            .then(options.postFilter || _.identity);
     }
 
     getChildContext() {
--- a/src/org-units-selector/OrgUnitsSelector.jsx
+++ b/src/org-units-selector/OrgUnitsSelector.jsx
@@ -1,5 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
+import _ from "lodash";
 
 import Card from "material-ui/Card/Card";
 import CardText from "material-ui/Card/CardText";
@@ -21,6 +22,8 @@ export default class OrgUnitsSelector extends React.Component {
         onChange: PropTypes.func.isRequired,
         selected: PropTypes.arrayOf(PropTypes.string).isRequired,
         levels: PropTypes.arrayOf(PropTypes.number),
+        rootIds: PropTypes.arrayOf(PropTypes.string),
+        listParams: PropTypes.object,
         controls: PropTypes.shape({
             filterByLevel: PropTypes.bool,
             filterByGroup: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -86,6 +89,46 @@ export default class OrgUnitsSelector extends React.Component {
         });
     }
 
+    getRoots({ filter }) {
```
SferaDev: [Proposal] [Untested]

```
    getRoots({ filter } = {}) {
        const { d2, listParams, rootIds } = this.props;

        const pageSize = !!rootIds ? 1000 : 10;
        const idFilter = !filter && rootIds ? `id:in:[${rootIds.join(",")}]` : null;

        const listOptions = _.omitBy(
            {
                paging: !!filter,
                pageSize: !!filter ? pageSize : null,
                level: !filter && !rootIds ? 1 : null,
                filter: filter || idFilter,
                fields: "id,displayName,path",
                ...listParams,
            },
            _.isNull
        );

        // We cannot both filter by name and check inclusion on rootIds on the same request, so
        // let's make a request filtering only by name and later check the rootIds in the response.
        const postFilter =
            filter && rootIds
                ? orgUnits =>
                      _(orgUnits)
                          .filter(orgUnit => rootIds.some(ouId => orgUnit.path.includes(ouId)))
                          .take(pageSize)
                          .value()
                : null;

        return d2.models.organisationUnits
            .list(listOptions)
            .then(collection => collection.toArray())
            .then(postFilter || _.identity);
    }
```

src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -76,7 +79,7 @@ export default class OrgUnitsSelector extends React.Component {
                       paging: false,
                       fields: "id,displayName",
                   }),
-            getDefaultRoots(props.d2),
+            this.getRoots({}),
```
SferaDev: Since the default call to getRoots is with "{}", wouldn't it make sense to change the signature to ```getRoots({ filter } = {})``` and leave here a ```this.getRoots();```.

src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -133,17 +176,8 @@ export default class OrgUnitsSelector extends React.Component {
...
-            : this.props.d2.models.organisationUnits
-                  .list({
-                      paging: true,
-                      pageSize: 10,
-                      fields: "id,displayName,path",
-                      filter: `displayName:ilike:${value}`,
-                  })
-                  .then(collection => collection.toArray()));
-
+        const opts = !value ? {} : { filter: `displayName:ilike:${value}` };
```
SferaDev: If the signature is changed this could also look like ```const opts = value ? { filter: `displayName:ilike:${value}` } : null;```.

src/org-units-selector/OrgUnitsSelector.jsx
SferaDev: Note: This is a mirror from the if/else logic in the commit, maybe I've misunderstood part of https://github.com/EyeSeeTea/vaccination-app/pull/237 and the conditions can be simplified more.

SferaDev (commented):
Code review looks good to me!

Feature-wise, I do not fully understand the implications of the change in vaccination-app and I could not properly test it.

src/org-units-selector/OrgUnitsSelector.jsx
tokland: This is a typical conditional code which can be implemented in these two ways: 1) as a truth table (as I did) or 2) yours, isolated code for each property. I know the truth table has some overlapping output, but my hope was that it was more clear this way, we have 4 cases and we define what to do on each of them.

src/org-units-selector/OrgUnitsSelector.jsx
tokland: Yes!

src/org-units-selector/OrgUnitsSelector.jsx
tokland: You mean `undefined`?

SferaDev: Yeah

src/org-units-selector/OrgUnitsSelector.jsx
SferaDev: Okay, looks good to me then.

src/org-units-selector/OrgUnitsSelector.jsx
tokland: Done!

PR #52, opened 2019-07-16
[ObjectsTable] Do not call formatter on null/undefined props

SferaDev:
### :memo: Implementation

- We are calling formatter even when value is null or undefined.

- This causes side-effects such as ```new Date()``` instead of ```undefined```

### :bookmark_tabs: Others

See the following example on https://github.com/EyeSeeTea/metadata-synchronization/pull/133.

Notice the details box and the lastExecuted property.

Before:

image

After:

image


tokland (approved):
Good catch!

Issue #7, opened 2019-02-22
Objects table this.state.pager is undefined


SferaDev:
```

:8081/static/js/0.chunk.js:87549 Uncaught TypeError: Cannot read property 'total' of undefined
    at calculatePageValue (:8081/static/js/0.chunk.js:87549)
    at ObjectsTable.getPaginationProps (:8081/static/js/0.chunk.js:87386)
    at ObjectsTable.render (:8081/static/js/0.chunk.js:87425)
    at finishClassComponent (:8081/static/js/0.chunk.js:217066)
    at updateClassComponent (:8081/static/js/0.chunk.js:217021)
    at beginWork (:8081/static/js/0.chunk.js:218011)
    at performUnitOfWork (:8081/static/js/0.chunk.js:222031)
    at workLoop (:8081/static/js/0.chunk.js:222072)
    at HTMLUnknownElement.callCallback (:8081/static/js/0.chunk.js:202094)
    at Object.invokeGuardedCallbackDev (:8081/static/js/0.chunk.js:202143)
    at invokeGuardedCallback (:8081/static/js/0.chunk.js:202197)
    at replayUnitOfWork (:8081/static/js/0.chunk.js:221255)
    at renderRoot (:8081/static/js/0.chunk.js:222185)
    at performWorkOnRoot (:8081/static/js/0.chunk.js:223109)
    at performWork (:8081/static/js/0.chunk.js:223019)
    at performSyncWork (:8081/static/js/0.chunk.js:222993)
    at requestWork (:8081/static/js/0.chunk.js:222848)
    at scheduleWork (:8081/static/js/0.chunk.js:222661)
    at Object.enqueueSetState (:8081/static/js/0.chunk.js:213515)
    at ObjectsTable.push../node_modules/react/cjs/react.development.js.Component.setState (:8081/static/js/0.chunk.js:232200)
    at ObjectsTable._callee$ (:8081/static/js/0.chunk.js:87347)
    at tryCatch (:8081/static/js/0.chunk.js:234237)
    at Generator.invoke [as _invoke] (:8081/static/js/0.chunk.js:234456)
    at Generator.prototype.(:8081/anonymous function) [as next] (http://localhost:8081/static/js/0.chunk.js:234290:21)
    at asyncGeneratorStep (:8081/static/js/0.chunk.js:87006)
    at _next (:8081/static/js/0.chunk.js:87028)
[...]

PR #57, opened 2019-07-26
[ObjectsTable] Improvements

### LINKED ISSUE #10: ObjectsTable improvements

tokland:
Closes #10 

High Priority
- [x] ObjectsTable: Hide checkboxes if there are no fields with multiple set to true.
- [x] Apply branch fix/calculate-page-value
- [x] ObjectsTable: We have `initialSorting` as a required prop, make it optional and assume first sortable column ASC is the default order.
- [x] Update readme for ObjectsTable with explanation regarding prop 'key' to update it.

Low Priority
- [x] Reduce marginTop by 10px, too much space.
- [x] Change model prop to optional (See https://github.com/EyeSeeTea/metadata-synchronization/issues/7 -> Remove model prop from instance-configurator)
- [x] Hide details panel if table refreshes otherwise it can be confusing for the user.

### DIFF
--- a/src/objects-table/ObjectsTable.js
+++ b/src/objects-table/ObjectsTable.js
@@ -24,7 +24,7 @@ class ObjectsTable extends React.Component {
         onButtonClick: PropTypes.func,
         pageSize: PropTypes.number.isRequired,
         model: PropTypes.object.isRequired,
-        initialSorting: PropTypes.array.isRequired, // [fieldName, "asc" | "desc"]
+        initialSorting: PropTypes.array, // [columnName: string, "asc" | "desc"]
         actions: PropTypes.arrayOf(
             PropTypes.shape({
                 name: PropTypes.string.isRequired,
@@ -45,7 +45,7 @@ class ObjectsTable extends React.Component {
                 style: PropTypes.object,
                 contents: PropTypes.element,
             })
-        ),
+        ).isRequired,
         /*  list: async function that returns paginated D2 objects.
 
             list(
@@ -125,13 +125,21 @@ class ObjectsTable extends React.Component {
         }));
     }
 
+    getDefaultSorting() {
+        const columnName = _(this.props.columns)
+            .map(column => (column.sortable ? column.name : null))
+            .compact()
+            .first();
+        return columnName ? [columnName, "asc"] : null;
+    }
+
     _getInitialState() {
         return {
             isLoading: true,
             page: 1,
             pager: { total: 0 },
             dataRows: [],
-            sorting: this.props.initialSorting,
+            sorting: this.props.initialSorting || this.getDefaultSorting(),
             searchValue: null,
             detailsObject: null,
             selection: new Set(this.props.initialSelection),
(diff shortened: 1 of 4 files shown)


SferaDev:
Found a "bug" with SyncRulesWizard in metadata-synchronization related to "Hide checkboxes if there are no fields with multiple set to true".

In the wizard we are using the multiple selection but without using ObjectsTable's actions, instead we use the Wizard pagination. Now the checkboxes are gone and that functionality is gone.

We could either add a multiple selection action and it would continue working but I guess that's not the perfect solution.

image

EDIT: 

Actually this happens for all synchronization pages too, as we rely on the FAB action instead of the row actions

image

src/objects-table/ObjectsTable.js
```diff
@@ -131,7 +144,7 @@ class ObjectsTable extends React.Component {
             page: 1,
             pager: { total: 0 },
             dataRows: [],
-            sorting: this.props.initialSorting,
+            sorting: this.props.initialSorting || this.getDefaultSorting(),
```
SferaDev, line 151: Tested. Works perfect!

src/objects-table/ObjectsTable.js
```diff
@@ -444,9 +460,9 @@ class ObjectsTable extends React.Component {
     }
 }
 
-function calculatePageValue(pager, defaultPerPage) {
-    const { total, pageCount, page, query } = pager;
-    const pageSize = query ? query.pageSize : defaultPerPage;
+function calculatePageValue(pager, pageSize) {
+    const { total, page } = pager;
+    const pageCount = Math.ceil(total / pageSize);
```
SferaDev, line 469: Tested. Works perfect!

src/objects-table/ObjectsTable.js
```diff
@@ -366,7 +382,7 @@ class ObjectsTable extends React.Component {
         const primaryAction = defaultAction ? defaultAction.fn : undefined;
 
         return (
-            <div>
+            <div style={styles.mainWrapper}>
```
SferaDev, line 389: Tested. Works perfect!

SferaDev (changes requested):
The only thing that blocks this PR is finding a wait for metadata-synchronization to force checkboxes without having a multi selection action

src/objects-table/ObjectsTable.js
tokland: Fixed: eb7c55dc

tokland:
Selection column problem fixed in eb7c55dc. 

PR for metadata-sync: https://github.com/EyeSeeTea/metadata-synchronization/pull/139

SferaDev (approved):

PR #61, opened 2019-08-07
[Wizard] Add optional prop warning and show above step description

tokland:
Adds optional prop warning and show above the step description (non-dismissable).

image


SferaDev (approved):
Work like a charm, tested on metadata-sync without issues

PR #69, opened 2019-10-22
bump sass version to make it compatible with node 12

adrianq:


SferaDev (approved):
Applied the branch and tested locally with metadata-sync, everything appears to be working properly after the version bump.

I've checked the release changelog and https://github.com/sass/node-sass/pull/2633, and the only changes are version version upgrades in their end.

The release was created 6 months ago so it is safe to assume we're good to go, also we will try to remove this dependency in future versions altogether.

tokland (approved):
:+1: Tested with skeleton-app and vaccination-app, it works fine. node v12.12.0

adrianq:
Thanks guys!

PR #70, opened 2019-10-22
Sync eslint with skeleton-app

SferaDev:


tokland (approved):
Yes, that's the best way, use extends. Added TS extensions.

SferaDev:
> Yes, that's the best way, use extends. Added TS extensions.

We don't have the extensions in skeleton-app. Without adding them it ignores TS files. I believe we should fix that there too.

PR #72, opened 2019-10-25
Add withElevation prop

xurxodev:
### :pushpin: References

* **related  PR:** https://github.com/EyeSeeTea/metadata-synchronization/pull/195

### :memo: Implementation

The goal of this PR is to create a new prop in OrgUnitsSelector to assign from outside if elevation must be visible or not.

OrgUnitsSelector component use this import `import Card from "material-ui/Card/Card"` and seems an old card component where elevation property not work. On this import `import Card from "material-ui/Core"` elevation property work.

To modify OrgUnitsSelector as little as possible, I have created a new prop withElevation with true value by default for maintaining current behaviour.

If withElevation is false I have added a new field boxShadow to styles.cardWide assigning "none" value.

### DIFF
--- a/src/org-units-selector/OrgUnitsSelector.jsx
+++ b/src/org-units-selector/OrgUnitsSelector.jsx
@@ -195,13 +195,11 @@ export default class OrgUnitsSelector extends React.Component {
         const getClass = root => `ou-root-${root.path.split("/").length - 1}`;
         const leftStyles = someControlsVisible ? styles.left : styles.leftFullWidth;
 
-        if (!withElevation) {
-            styles.cardWide.boxShadow = "none";
-        }
-
+        const cardWideStyle = withElevation ? styles.cardWide : {...styles.cardWide, boxShadow: "none"}
+ 
         return (
             <div>
-                <Card style={styles.cardWide}>
+                <Card style={cardWideStyle}>
                     <CardText style={styles.cardText}>
                         <div style={styles.searchBox}>
                             <SearchBox onChange={this.filterOrgUnits} />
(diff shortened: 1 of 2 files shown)


src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -185,14 +187,18 @@ export default class OrgUnitsSelector extends React.Component {
...
+        const { selected, controls, withElevation } = this.props;
         const { filterByLevel, filterByGroup, selectAll } = controls;
         const someControlsVisible = filterByLevel || filterByGroup || selectAll;
         const { renderOrgUnitSelectTitle: OrgUnitSelectTitle } = this;
         const initiallyExpanded = roots.length > 1 ? [] : roots.map(ou => ou.path);
         const getClass = root => `ou-root-${root.path.split("/").length - 1}`;
         const leftStyles = someControlsVisible ? styles.left : styles.leftFullWidth;
 
+        if (!withElevation) {
+            styles.cardWide.boxShadow = "none";
```
tokland: As a rule of thumb, we tend to write immutable code. Here, there is also a practical reason to do so, as `styles.cardWide` is used later as a prop. Classical approach:

```
const cardWideStyle = withElevation ? styles.cardWide : {...styles.cardWide, boxShadow: "none"}
...
 <Card style={cardWideStyle}>
```

xurxodev: ok, fixed!

tokland:
Merged into #83, thanks @xurxodev !

PR #81, opened 2019-10-30
[MultiSelector] Allow search box filtering

SferaDev:
### Implementation

- Add a new optional prop ``searchFilterLabel``
  - If defined a search box will appear and will filter in both sides
  - If not defined the search box will be hidden
- We rely on the filtering of the ``GroupEditorWithOrdering``
- We use old ``material-ui`` because we use everywhere the same search box and to maintain the look and feel, once we update everywhere to v4 we should take care of this component too


tokland:
Integrated in #83, closing this PR.

Issue #74, opened 2019-10-29
Upgrade to material v4


tokland:
- [x] Upgrade material v3 -> v4.5.2
- [x] Upgrade material v0 -> v4.5.2 (own components)
- [x] Upgrade material v0 -> v4.5.2 (d2-ui orgUnits/groupEditor -> components: RaisedButton, LinearProgress, Paper)
- [x] Integrate https://github.com/EyeSeeTea/d2-ui-components/pull/81
- [x] Integrate https://github.com/EyeSeeTea/d2-ui-components/pull/72
- [x] Multiselector orange

tokland:
Moving security PR to another issues, it makes more to run it after all those migration PRs have been merged to development, in a separate PR.

PR #82, opened 2019-11-06
Build new DataTable into TS

SferaDev:
## Important

- I had to update TS, babel and prettier because we are already using newer semantics introduced in TS 3.4 and 3.5 (such as ``as const``).

- I had to update ESLint config because it was failing in TS syntax (such as ``T extends``). I used the one of the new skeleton app for consistency.

- ``NewDataTable``, ``NewDataTableProps``, `` NewObjectsTable`` and ``NewObjectsTableProps`` should be properly renamed once we remove the old components (possible transition is to rename the old ones to ``OldDataTable`` and ``OldObjectsTable`` first and remove later).

- I've already introduced the  new syntax for export as used in ``d2-api`` (Example ``export * from`` ).

- During the review some parts of the original message have been already solved but the message itself has not been updated yet.

### :pushpin: References

* **DO NOT MERGE:** This has been developed here but meant to be moved to ``d2-ui-components`` once we have material v4 there.

* **Related issue:** https://github.com/EyeSeeTea/d2-ui-components/issues/68

#### File arrangement and structure

The granularity to divide components in different files is ``medium`` sized meaning only conceptually equivalent elements have been grouped in common files.

- utils
    - filtering: Utilities to allow search filters (this is only used in ObjectsTable and could be potentially moved there).
    - formatting: "Smart formatter" that allows correctly printing an element if it's an array of elements, d2-like object that provides displayName, name or id, date or moment or url. If nothing matches defaults to the element provided.
    - selection: Set of utility methods used to handle the selection of rows and actions. Includes the helper methods to create table notifications (select in all pages).
    - sorting: Set of utility methods to perform the sorting (this should be revisited very soon to improve the algorithm and allow automated sorting of d2-like elements).
- ActionButton: Floating action button used in ObjectsTable (refactored into TS from d2-ui-components).
- ContextualMenu: Popover menu printed when an action is requested.
- DataTable: Main component code.
- DataTableBody: Component similar to ``tbody``.
- DataTableHeader: Component similar to ``thead``.
- DataTableNotifications: Component refactored from d2-ui-components to show messages/warning/error on-demand (for example, select in multiple pages)
- DataTablePagination: Wrapper of the material-ui TablePagination.
- DetailsBox: Component refactored from d2-ui-components to show a details box.
- ObjectsTable: DataTable wrapper that adds 1) searchbox, 2) details box and 3) action button.
- types: Type definition file for generic types (prop types are in each component file)

#### Contextual actions

Checkboxes are hidden if no multiple action has been defined or ``forceSelectionColumn`` is set to false. Similarly if ``forceSelectionColumn`` is set to true checkboxes are always shown even though there are no multiple actions.

image

Icons are optional by default and expects a ``ReactNode``. In ObjectsTable the details icon is injected if no icon is provided. Similarly the details onClick action has injected the method to open the panel but still calls the parent onClick receiver.

image

Note that in the old ObjectsTable/DataTable we returned either an object or an array of objects for the onClick, now we always return an array of objects and it's safe to assume that for single actions the row will be located at array[0]. The code will never call the onClick if there are no selected rows, however I could not find a proper way to type that behavior.

#### Controlled/uncontrolled behavior

This is still a work-in-progress, the behavior in this pull request should be considered a first option. Right now we expose three optional props that act as uncontrolled props: ``pagination``, ``selection`` and ``sorting``.

The use cases for this uncontrolled behavior are the following:

- Pagination: Most common scenario. Allow pagination to next/prev pages not yet provided to the table, as in server-side pagination.

- Selection: Used in apps like metadata-sync. Allow selecting elements programmatically due to events not available inside the table.

- Sorting: Probably the least used scenario. Sort by a custom comparator (priority sorting, nested sorting, etc.). This could've been implemented with just a comparator func but for consistency I applied the same pattern as before.

How it works? If you pass the ``pagination``, ``selection`` or ``sorting`` props the table ignores the auto-computed value and just uses the provided one. However the internals still work and the events (change page, change selection, change sorting) are notified to parent through a ``onChange`` property method.

If only a partial-uncontrolled behavior is desired we also offer ``initialPagination``, ``initialSelection`` and ``initialSorting`` to preload the state instead of controlling it from the parent.

#### Positioning

Since we want to make the DataTable as customizable as possible we have added two props that allows inserting components either in the Toolbar (where the Pagination is located) and next to the DataTable itself (DetailsBox for example).

Both properties expect either a single ReactNode or a collection of them. Since they are rendered inside a ``flex`` layout, you can position them anywhere in the screen with the ``order`` property.

Allowing the application to do things like the following:

image

image

#### Table notifications

Previously we only included notifications related to the selection. We now allow to provide external customizable notifications.

image

#### Selection in all pages

The behavior is the same as we had in the original ObjectsTable, however now we do not perform two calls to a listing method with paging false as we previously did.

Now we control it through a prop called ``idsForSelectInAllPages`` (could be easily renamed) that has three different states:

- undefined: Default behavior. Auto-computed select in all pages according to the ``objects`` passed down as rows.
- string[]: Array of ``id`` strings that represent exisiting (or plausible) rows. We do not verify the objects are in rows as they could be in paginated rows not available to the table.
- null: Disables the select in all pages behavior, the message to select everywhere never appears. However if there are selected items in other pages or hidden trough filtering the other-pages-selection still appears.

image

### :fire: Is there anything the reviewer should know to test it?

- Beware this is a huge refactor, here be dragons!

### :bookmark_tabs: Things to revisit in future PRs

- Implement the filter columns feature and allow a way to force show/hide. Right now the icon is visible to keep in mind the TODO. Feel free to hide under an always false condition until ready. The same button could have multiple external actions (download as csv, etc) and show a wheel instead. To be discussed as-per the requirements of Data Sync and Samaritan's.

- Select in all pages: I just copied the same code we had in the previous ObjectsTable, this could be probably improved.

- Default sorting for decimal numbers does not work properly

- We are not doing any special memoization or optimization, as suggested by the react core team, it's better to create the components first and do a performance analysis and only tune the components that re-render the most or take a lot of time to re-render.

- Table notifications: Remove ``newSelection`` prop and properly handle ``onClick`` action.

- Either in d2-ui-components or d2-api we could create both a DataTable and ObjectsTable wrapper that manages the uncontrolled pagination and updates the sorting with server side object loading.

- Search box does not allow customizing ``hintText`` and ``initialValue`` in d2-ui-components. We should add the prop there since we pass the customization here. Probably refactoring it into TS altogether. Keep in mind that this is still material v0.


tokland (approved):
Commented with @SferaDev and committed some changes. Good job!

SferaDev:
Yeah, there are some things to improve but overall we have a good start point. Once we begin using it in real-world scenarios we might change some implementation details to make it completely usable.

SferaDev:
@adrianq Renamed and folders moved accordingly. Also merged with latest upstream changes of #74

adrianq:
For the record, the list of things to improve that can be found at the bottom of the PR description has been caught at https://github.com/EyeSeeTea/d2-ui-components/issues/85

PR #89, opened 2019-11-11
Feature/ou-level-wizard #21

laurapareja:
### :pushpin: References

* **Issue:** Closes#21 Project Monitoring
Closes #89  

* **d2-ui-components**: Requires #89 

### :memo: Implementation
- Include `selectableLevels` prop in `OrgUnitTree` (by default is all selectable). 
- `selectableLevel` is an Array 
- the biggest number of `selectableLevel` hasn't children.
- Include `typeInput` prop in `OrgUnitTree` (by default is checkbox)

### :art: Screenshots
**Example with one `selectableLevels`**:
``` const selectableLevels = [3];```

image

**Example with `selectableLevels`**:
``` const selectableLevels = [1, 3];```

In this case, level 3 is the last selectable level so it doesn't have any children 

image

**Example without `selectableLevel` and `typeInput` props**, default OrgUnitTree

image

### CODE REVIEW

File: src/org-unit-tree/OrgUnitTree.component.js [hunk 1]
```diff
@@ -234,6 +244,14 @@ class OrgUnitTree extends React.Component {
             this.props.onChangeCurrentRoot(currentOu);
         };
 
+        const handleTypeLabel = () => {
+            if (selectableLevel !== undefined) {
+                return "radio";
```
tokland: I forgot about this: whether a radio or checkbox is to be used does not depend on `selectableLevel`, it should be a separate prop (if we really want to use a radio input here).

laurapareja: In the last commit, you can see another prop `typeInput`. By default, without this prop, it would be checkbox.

PR #93, opened 2019-11-18
Update return type of date picker onChange

SferaDev:


SferaDev:
Will do this small change in the other PR for simplicity

PR #94, opened 2019-11-19
[DataTable] Render children rows and allow selection

SferaDev:
### :memo: Implementation

- Allow rendering children rows based on a new prop that allows

### :art: Screenshots

image

### :fire: Is there anything the reviewer should know to test it?

- N-levels of children.
    - Even though theoretically supported it has not been tested

- Selection messages
    - There is a known issue that I had no time to solve yet with the "selection in other pages" message.
    - Right now it appears one message telling that the children items are not in the current page (and that is true because the children rows are not normal rows).
    - The solution is not trivial and I'd like to leave it for next week in another PR.

### :bookmark_tabs: Others

- @tokland I initially began using ``(keyof T)[]`` instead of ``string[]`` but it added a level of complexity with nested fields. I could not make it work and I believe it has to do with how the inference works in ``d2-api``. I would not worry about it now but maybe we should take a look without rush.


src/data-table/DataTable.tsx
```diff
@@ -65,6 +65,7 @@ export interface DataTableProps<T extends ReferenceObject> {
     sideComponents?: ReactNode; // Portal to right-most of the Data Table
     ids?: string[]; // Enables selection in all pages (disabled with [])
     loading?: boolean;
+    childrenTags?: string[];
```
tokland, line 68: Maybe `childrenKeys`?

PR #98, opened 2019-11-26
Hide stepper if only one step is provided

adrianq:


adrianq:
We need a PR for this, right @SferaDev?

PR #100, opened 2019-12-11
[DatePicker] Allow all mui component props

tokland:
Instead of hardcoding the props we accept, allow all props for the mui date picker component.

Required by https://github.com/EyeSeeTea/project-monitoring-app/pull/88


adrianq:
@tokland You labelled this release as 1.1.0-beta.1 However, keep in mind we haven't released 1.0.0 as we are waiting for an app release. This is not a problem for now but let's try not to skip 1.0.0 in the stable channel

Issue #97, opened 2019-11-20
[DataTable] Improvements and fixes


tokland:
- [x] Maintain table order for children rows

- [x] Pin to 25 rows per page by default

- [x] Fix bug with pre-calculated pagination

- [x] Move to page 1 when search changes

PR #101, opened 2019-12-17
[DataTable] Recalculate pagination programmatically

SferaDev:
Since we were forcing the total to be initialized on first render we took ``objects.length`` only once, if the application updates the objects collection we still had the original value.

Instead set it to undefined on initialPagination and calculate the value every time. Note that we would not calculate it if the pagination is controlled by the app.

Also in uncontrolled mode fix the reset pagination to page 1 when sorting changes and make default page size 25. (#97)

### To test

Option 1:
- https://github.com/EyeSeeTea/metadata-synchronization/pull/252
- Events sync wizard, "Events" step.

Option 2:

- Simple ObjectsTable with rows initially ``[]`` and then populated without controlling the pagination.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
         "type": "git",
         "url": "git+https://github.com/eyeseetea/d2-ui-components.git"
     },
-    "version": "1.0.0-beta.7",
+    "version": "1.0.0-beta.8",
     "main": "index.js",
     "types": "index.d.ts",
     "peerDependencies": {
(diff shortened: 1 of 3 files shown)


tokland (commented):
The code looks good to me and it works with project-monitoring-app. Just a caveat: I had to change a `initialState={{ pagination }}` I had, otherwise the pagination wouldn't change when using the search bar. The new value is: `const pagination: TablePagination = { pageSize: 10, page: 1, total: 0 };`. Maybe `total` should be optional so I am not forced to send this hacky `total: 0` or should I send the props in another way? I want to set a pageSize/page, but not the total, which should be calculated.

SferaDev:
@adrianq I've updated the PR with a proposal to be reviewed.

There's no rush merging this PR and it can wait until next week.

However you might need to pick this locally to test ``metadata-sync``.

adrianq:
@tokland we are merging this now but if you can please confirm the latest commit from @SferaDev fixes your issue it would be great!

PR #104, opened 2019-12-31
[DataTable] Add basic support for column hiding

SferaDev:
### :memo: Implementation

- [FIX] Sort columns with ``getValue()``
- [BREAKING] TableColumn's ``name`` is now ``string`` instead of ``keyof T``. There might be columns that do not reference any property of T.
- [FEATURE] Allow to hide columns with ``hide === true``
- [FEATURE] Allow to show/hide columns on-demand through a dialog (can be disabled with ``hideColumnVisibilityOptions``

### :art: Screenshots

image
image

### DIFF
--- a/src/data-table/DataTableBody.tsx
+++ b/src/data-table/DataTableBody.tsx
@@ -40,6 +40,7 @@ const rotateIconStyle = (isOpen: boolean) => ({
 export interface DataTableBodyProps<T extends ReferenceObject> {
     rows: T[];
     columns: TableColumn<T>[];
+    visibleColumns: (keyof T)[];
     sorting: TableSorting<T>;
     primaryAction?: TableAction<T>;
     selected: string[];
@@ -54,6 +55,7 @@ export function DataTableBody<T extends ReferenceObject>(props: DataTableBodyPro
     const {
         rows,
         columns,
+        visibleColumns,
         sorting,
         primaryAction,
         selected,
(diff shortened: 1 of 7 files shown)


adrianq:
@SferaDev It works nicely from the project monitoring app. I will wait for @tokland review but no defect from a functional point of view.

src/data-table/DataTableBody.tsx
```diff
@@ -54,41 +55,45 @@ export function DataTableBody<T extends ReferenceObject>(props: DataTableBodyPro
...
-        if (event.type === "contextmenu") {
-            event.preventDefault();
-            contextualAction(row, event);
-        } else if (enableMultipleAction && (isEventCtrlClick(event) || isCheckboxClick)) {
-            onChange(updateSelection(selected, row));
-        } else if (primaryAction && primaryAction.onClick) {
-            primaryAction.onClick([row]);
-        }
-    };
+    const isSelected = (name: string) => selected.indexOf(name) !== -1;
```
tokland, line 72: [very minor] `Array.includes` or `_.includes` is more declarative.

src/data-table/types.ts
```diff
@@ -8,9 +8,10 @@ export interface TableObject extends ReferenceObject {
 }
 
 export interface TableColumn<T extends ReferenceObject> {
-    name: keyof T;
+    name: string;
```
tokland, line 11: On the next meeting, we can discuss on this, I was thinking if the T object should always contain the column (if it's a transform column, then transform it before). But for now it's ok, as its widening the type, not narrowing it, so it does not break anything.

tokland (approved):
About the UI: 1) I guess we simplified by not allowing re-ordering of fields like in user-app 2) Aren't the checks a bit too separate? About the code, just two minor comments, the second one is something to see in the future, no problem to commit like this.

adrianq:
@SferaDev after discussing with @tokland I will go ahead and merge this PR. None of tokland's comments is too critical. If you wish you can add any change in another PR... Thanks guys!

PR #108, opened 2020-01-03
[ObjectsTable] Refactor sorting and fix minor issues

### LINKED ISSUE #107: Type TableColumn as keyof T
As we had before, use keyof T instead of string. That means that virtual fields must come already transformed in the row from the app.

https://github.com/EyeSeeTea/d2-ui-components/commit/aa1ce1f61eb35d88f5bb24475e82385064c0b98f

tokland:
Closes #107


SferaDev (approved):
Tested on metadata-sync and everything works fine.

Thanks!

PR #109, opened 2020-01-08
[ObjectsTable] Allow sending selection customization from application

SferaDev:
### 🎯 Objectives

- Allow excluding some rows from parent-children rows

### :memo: Implementation

- Change selection type from ``string[]`` to ``TableSelection[]``
- ``TableSelection`` allows to provide ``checked``, ``indetermindate`` and ``icon`` props to the row CheckBox API.
- Remove old ad-hoc children auto-selection

### :art: Screenshots

image

### DIFF
--- a/src/data-table/utils/selection.tsx
+++ b/src/data-table/utils/selection.tsx
@@ -35,9 +35,7 @@ export function getActionRows<T extends ReferenceObject>(
     selection: TableSelection[]
 ) {
     const isRowInSelection = _.some(selection, { id: selectedRow.id });
-    const selectedRows = _(allRows)
-        .intersectionBy(selection, "id")
-        .value();
+    const selectedRows = _.intersectionBy(allRows, selection, "id");
 
     return isRowInSelection ? selectedRows : [selectedRow];
 }
--- a/src/data-table/utils/selection.tsx
+++ b/src/data-table/utils/selection.tsx
@@ -34,10 +34,9 @@ export function getActionRows<T extends ReferenceObject>(
     allRows: T[],
     selection: TableSelection[]
 ) {
-    const isRowInSelection = !!_.find(selection, { id: selectedRow.id });
-    const selectedRows = _(selection)
-        .map(({ id }) => _.find(allRows, ["id", id]))
-        .compact()
+    const isRowInSelection = _.some(selection, { id: selectedRow.id });
+    const selectedRows = _(allRows)
+        .intersectionBy(selection, "id")
         .value();
 
     return isRowInSelection ? selectedRows : [selectedRow];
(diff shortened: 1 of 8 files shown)


src/data-table/utils/selection.tsx
```diff
@@ -1,39 +1,46 @@
...
 }
 
-export function getActionRows<T extends ReferenceObject>(row: T, rows: T[], selection: string[]) {
-    const rowInSelection = !!selection.find(id => row.id === id);
+export function getActionRows<T extends ReferenceObject>(
+    selectedRow: T,
+    allRows: T[],
+    selection: TableSelection[]
+) {
+    const isRowInSelection = !!_.find(selection, { id: selectedRow.id });
```
tokland: minor: `!!_.find` -> `_.some`. 

That `{ id: selectedRow.id }` matcher is cool, but I am afraid TS does not type-check it.

src/data-table/utils/selection.tsx
```diff
@@ -1,39 +1,46 @@
...
+    const selectedRows = _(selection)
+        .map(({ id }) => _.find(allRows, ["id", id]))
```
tokland: A `find` of N nested in a `map` of M is _O(n*m)_, which can get quite large on real scenarios. Typical solution: process the inner data structure so the predicate is O(1).

src/old-objects-table/ObjectsTable.js
```diff
@@ -315,7 +315,10 @@ class ObjectsTable extends React.Component {
         if (_.isEmpty(dataRows) || disableMultiplePageSelection) return [];
 
         const allSelected = selection.size === pager.total;
-        const selectionInOtherPages = _.difference([...selection], dataRows.map(dr => dr.id));
+        const selectionInOtherPages = _.difference(
```
tokland, line 318: minor: This gets converted to one line when running "yarn prettify"

tokland (commented):
No problems with with project-monitoring, I had just to change the selection array. Some comments regarding the code, not very important:

src/data-table/utils/selection.tsx
SferaDev: Yeah I noticed that too. I am slowly switching to ``_.find(array, ["id", id])`` but I find the ``_.matchesProperty`` shorthand much more readable.

SferaDev:
@adrianq I will update the PR with these comments during the night. But if you're testing them for metadata-synchronization feel free to merge this as-is and I will add the changes to the next release. Your call!

src/data-table/utils/selection.tsx
SferaDev: I did face the type-check error sometimes but here it does indeed properly work :o

image

src/data-table/utils/selection.tsx
SferaDev: Simplified with:

```ts
const selectedRows = _.intersectionBy(allRows, selection, "id");
```

Thanks @tokland!

tokland (approved):
All good!

adrianq:
All good here! Thanks @SferaDev

PR #111, opened 2020-01-13
[OrgUnitSelector] Visual updates

SferaDev:
Since @tokland and I will be using this component for two projects at the same time I believe we could use this PR to make all the changes together.

This works as-is for "metadata-sync" and theoretically none of these changes should visually break anything for existing apps.

However the code still has a lot of hardcoded styling that makes it really hard to control the component from the application side.

Example of how it visually looks in metadata-sync right now, as reference for what we currently need in that app:

image
image
image
image
image

Note: The pictures do not necessarily represent actual final features, just UI positioning.


src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -313,7 +335,6 @@ const styles = {
         display: "inline-block",
```
tokland: I had to remove this `display: "inline-block"` to see it as I needed. This way, the component uses visually the full width, and now the parent div can decide the width. This works fine for bulk-load and project-monitoring-app. Should be tested with filters too.

SferaDev: Okay, I will try to remove it then and see how it looks like on metadata-synchronization too.

tokland (approved):
Great, it has all the changes we discussed. It looks good with project-monitoring-app and bulk-load.

adrianq:
Thanks guys for the work! Tested in bulk, project-monitoring and metadata. It seems to be working fine. I have also released a new version of d2-ui-comp in the beta channel and updated the relevant app together with @tokland

PR #114, opened 2020-01-22
DialogButton/ConfirmationDialog: Allow react nodes as title/description

tokland:
Required by https://github.com/EyeSeeTea/project-monitoring-app/pull/125

### DIFF
--- a/src/confirmation-dialog/ConfirmationDialog.jsx
+++ b/src/confirmation-dialog/ConfirmationDialog.jsx
@@ -6,8 +6,8 @@ import i18n from "../utils/i18n";
 class ConfirmationDialog extends React.Component {
     static propTypes = {
         isOpen: PropTypes.bool.isRequired,
-        title: PropTypes.string.isRequired,
-        description: PropTypes.string,
+        title: PropTypes.node.isRequired,
+        description: PropTypes.node,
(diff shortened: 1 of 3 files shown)


src/confirmation-dialog/ConfirmationDialog.jsx
```diff
@@ -61,4 +55,12 @@ class ConfirmationDialog extends React.Component {
     }
 }
 
+function renderNode(node) {
+    if (typeof node === "string") {
```
SferaDev, line 59: When I needed to display non-string nodes as part of the "description" I normally sent them as ``children`` wrapped with a ``DialogContent``.

The change as-is looks good to me, but if we want to avoid the ``typeof`` checking for ``description``, that could still be an alternative.

tokland: >  I normally sent them as children

Exactly my thoughts, and in fact we are already rendering the children in a DialogContent. The problem is that we were already doing a post-processing on the description (split to use html paragraphs), so I didn't found another way of not breaking existing code.

SferaDev (approved):
Works fine in metadata-sync: tested sending single-line, multi-line and nodes to both properties and everything works as expected.

PR #116, opened 2020-01-22
Feature/disable scroll lock context menu

tokland:
Required by https://github.com/EyeSeeTea/project-monitoring-app/pull/126

When the context menu is open, material-ui made some changes to the html/body to disable the overflow scrollbar. This moved some parts of the page to the right (ugly effect). 

See https://github.com/mui-org/material-ui/issues/17353

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
         "type": "git",
         "url": "git+https://github.com/eyeseetea/d2-ui-components.git"
     },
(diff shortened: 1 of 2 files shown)


SferaDev (approved):
LGTM. Could not reproduce the bug but the change does not seem to introduce any visual effect on metadata-sync.

PR #117, opened 2020-01-22
[OrgUnitTree] Remove d2 dependency from OrgUnit components

SferaDev:
* **NOTE:** This is a BREAKING change. Version should be bumped one ``minor`` semver. If we do not want to do it breaking, we should duplicate the components and offer the ``OldOrgUnitSelector`` which might be an overkill.

Removed d2 depenendency from ``OrgUnitTree`` and ``OrgUnitSelector``.

Touched the minimum code to migrate from ``d2`` to ``d2-api``.

image

### DIFF
--- a/src/org-unit-select/OrgUnitSelectAll.component.js
+++ b/src/org-unit-select/OrgUnitSelectAll.component.js
@@ -48,10 +48,12 @@ class OrgUnitSelectAll extends React.Component {
         } else {
             this.setState({ loading: true });
 
-            this.context.d2.models.organisationUnits
-                .list({ fields: "id,path", paging: false })
+            this.context.api.models.organisationUnits
+                .get({ fields: { id: true, path: true }, paging: false })
+                .getData()
+                .then(({ objects }) => objects)
                 .then(orgUnits => {
-                    const ous = orgUnits.toArray().map(ou => ou.path);
+                    const ous = orgUnits.map(ou => ou.path);
                     this.setState({
                         cache: ous,
                         loading: false,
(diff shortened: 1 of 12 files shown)


src/org-unit-select/OrgUnitSelectAll.component.js
```diff
@@ -67,12 +69,15 @@ class OrgUnitSelectAll extends React.Component {
...
     getDescendantOrgUnits() {
-        return this.context.d2.models.organisationUnits.list({
-            root: this.props.currentRoot.id,
-            paging: false,
-            includeDescendants: true,
-            fields: "id,path",
-        });
+        return this.context.api.models.organisationUnits
+            .get({
+                root: this.props.currentRoot.id,
```
tokland, line 74: I thought this (using a non-standard _model.get_ field) works here in JS because of the lack of typing. In fact, it also works in TS because the machinery necessary to make the get pagination/unpaginated `get` work, allows passing any extension of the options. That comes handy here, but we would to revisit it properly in d2-api. Issue created: https://github.com/EyeSeeTea/d2-api/issues/19

src/org-unit-select/OrgUnitSelectByGroup.component.js
```diff
@@ -65,17 +69,23 @@ class OrgUnitSelectByGroup extends React.Component {
...
-                    .then(orgUnits => {
-                        log.debug(`Loaded ${orgUnits.length} org units for group ${groupId}`);
+                const { api } = this.context;
+                api.models.organisationUnitGroups
+                    .get({
+                        fields: { organisationUnits: { id: true, path: true } },
+                        filter: { id: { eq: groupId } },
+                    })
+                    .getData()
+                    .then(({ objects }) => _.first(objects) || {})
```
tokland, line 79: Added explicit import lodash [DONE]

src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -52,45 +46,48 @@ export default class OrgUnitsSelector extends React.Component {
...
             levels: null,
             roots: null,
             groups: null,
             currentRoot: null,
         };
         this.contentsStyle = { ...styles.contents, height: props.height };
+    }
 
+    componentDidMount() {
+        const { props } = this;
```
tokland, line 66: I moved this code from constructor, I was getting a warning on the console about calls to setState on non-mounted components. The original dhis2 code was wrong, no request should be made in a component constructor. [DONE]

tokland (approved):
Works nicely! Tested in project-monitoring-app. I just commited some minor changes.

adrianq:
@SferaDev I have updated the [release PR title](https://github.com/EyeSeeTea/d2-ui-components/pull/113) to 1.1.0 so we keep in mind this breaking change. Mental note for me we should use ~ instead of ^ for the releases. Not sure if I always do it this way

PR #119, opened 2020-01-29
Add disable prop to TableSelection

tokland:
Fow now, the `disabled` property is used only for the checkbox. Should we also style the row?

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
         "type": "git",
         "url": "git+https://github.com/eyeseetea/d2-ui-components.git"
     },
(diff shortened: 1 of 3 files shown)


tokland:
Simply adding a prop `disable` is not enough. A value `tableSelection` only makes sense when the row is selected, we want a general property of the row, also when not selected. 

Proposal: add property `selectable?: boolean` to `ReferenceObject`, and use it to determine if that row can be selected. Also, add background color to disabled rows.

https://slack-files.com/T8G8SCDGX-FTADXSP7W-11219ed10e

PR #123, opened 2020-02-04
Avoid infinite re-rendering when using inside a useEffect

SferaDev:
# Implementation

- Store context inside a ref and return the immutable value
- Expose state and ``openSnackbar``/``closeSnackbar`` methods
- Unify code between ``useLoading`` and ``useSnackbar``

### DIFF
--- a/src/snackbar/types.ts
+++ b/src/snackbar/types.ts
@@ -7,15 +7,14 @@ export interface SnackbarOptions {
     autoHideDuration?: number | null;
 }
 
-export interface SnackbarState extends SnackbarOptions {
-    success: (message: string, options?: Partial<SnackbarOptions>) => void;
-    info: (message: string, options?: Partial<SnackbarOptions>) => void;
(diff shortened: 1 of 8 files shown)


src/snackbar/types.ts
```diff
@@ -7,21 +7,15 @@ export interface SnackbarOptions {
     autoHideDuration?: number | null;
 }
 
-export interface SnackbarState {
+export interface SnackbarState extends SnackbarOptions {
+    success: (message: string, options?: Partial<SnackbarOptions>) => void;
```
tokland: Now that we are at it, can we create a type `(message: string, options?: Partial<SnackbarOptions>) => void` to DRY the interface?

tokland (approved):
I see, useRef does the trick. Actually, it makes more sense than useMemo in this case. Ok by me.

tokland (approved):
Working in project-monitoring-app

PR #131, opened 2020-02-24
[DialogButton] Pass rest props

tokland:
Required by https://github.com/EyeSeeTea/project-monitoring-app/pull/190


tokland:
@SferaDev Codes looks good, it builds and works in project-monitoring-app, so :+1: 

(Not sure about the README containing the webpack file, we can link to the repo itself. In any case, that's for another PR)

PR #132, opened 2020-02-25
[DataTable] Add global actions

SferaDev:
- Required by:  https://github.com/EyeSeeTea/metadata-synchronization/pull/386

image

### DIFF
--- a/src/data-table/ContextualMenu.tsx
+++ b/src/data-table/ContextualMenu.tsx
@@ -64,10 +64,6 @@ export function ContextualMenu<T extends ReferenceObject>(props: ContextualMenuP
                 left: positionLeft,
                 top: positionTop,
             }}
-            anchorOrigin={{
-                vertical: "center",
-                horizontal: "center",
-            }}
             onClose={onClose}
         >
             {actions.map(action => (
--- a/src/data-table/DataTableBody.tsx
+++ b/src/data-table/DataTableBody.tsx
@@ -78,6 +78,7 @@ export function DataTableBody<T extends ReferenceObject>(props: DataTableBodyPro
         const primaryAction = _(availableActions).find({ primary: true }) || defaultAction;
 
         const contextualAction = (event: MouseEvent<unknown>) => {
+            if (isEventCtrlClick(event)) return;
             event.stopPropagation();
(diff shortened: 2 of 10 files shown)


src/data-table/ContextualMenu.tsx
```diff
@@ -64,10 +64,6 @@ export function ContextualMenu<T extends ReferenceObject>(props: ContextualMenuP
                 left: positionLeft,
                 top: positionTop,
             }}
-            anchorOrigin={{
-                vertical: "center",
-                horizontal: "center",
-            }}
```
SferaDev, line 70: Removed this because it is ignored when working with ``anchorPosition`` (see https://material-ui.com/api/popover)

src/data-table/DataTableBody.tsx
```diff
@@ -78,6 +78,7 @@ export function DataTableBody<T extends ReferenceObject>(props: DataTableBodyPro
         const primaryAction = _(availableActions).find({ primary: true }) || defaultAction;
 
         const contextualAction = (event: MouseEvent<unknown>) => {
+            if (isEventCtrlClick(event)) return;
```
SferaDev, line 81: This fixes the control-click issue we have when working with Cypress FD recorder

tokland (approved):
Code looks good, tested in feature/tracker-mapping, worked nicely.

I tested it in project-monitoring and I've had some type errors (date picker, wizard validations), but they are not related to this PR, but previous PRs already merged in DEV, I've open an issue there whenever we want to use the main branch:

https://github.com/EyeSeeTea/project-monitoring-app/issues/197

SferaDev:
> I tested it in project-monitoring and I've had some type errors (date picker, wizard validations), but they are not related to this PR, but previous PRs already merged in DEV, I've open an issue there whenever we want to use the main branch:

Yeah, since now we expose the "correct" typings, TS might point some issues in the application side. However if you find some incoherence in the typings we should fix it here (optional/mandatory parameters or types).

tokland:
> Yeah, since now we expose the "correct" typings, TS might point some issues in the application side.

Yep, at first glance that was it, incorrect types on the app (when there is no compiler to enforce them...)

PR #133, opened 2020-02-27
[DataTable] Allow switching left/right click

tokland:
Required by:

https://github.com/EyeSeeTea/project-monitoring-app/issues/174
https://github.com/EyeSeeTea/project-monitoring-app/issues/179


src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -99,7 +99,7 @@ export default class OrgUnitsSelector extends React.Component {
     }
 
     getRoots({ filter } = {}) {
-        const { api, listParams, rootIds } = this.props;
+        const { api, listParams, rootIds, selectableLevels } = this.props;
```
SferaDev, line 102: I guess ``selectableLevels`` comes from d2-ui, anyway since it's not a disable but hiding operartion, would it make sense something like ``visibleLevels``?

SferaDev (approved):
Everything works like a charm on ``metadata-sync``!

src/org-units-selector/OrgUnitsSelector.jsx
tokland: Indeed, we saw the naming/logic is confusing but were in a hurry, so I opened a separate issue to fix it somewhere in time: https://github.com/EyeSeeTea/d2-ui-components/issues/134

SferaDev: Great then

PR #139, opened 2020-03-04
[SearchBox] Fix flickering due to debounce

### LINKED ISSUE #408: Searchbox flickering/ Cypress inestability
https://github.com/cypress-io/cypress/issues/5480

SferaDev:
* **Issue:** Closes https://github.com/EyeSeeTea/metadata-synchronization/issues/408

- Changed to _.debounce (we can remove the ``throttle-debounce`` dependency as soon as we deprecate the old objects table).

### DIFF
--- a/src/data-table/ObjectsTable.tsx
+++ b/src/data-table/ObjectsTable.tsx
@@ -1,4 +1,4 @@
-import React, { useState, ReactNode, MouseEvent } from "react";
+import React, { useState, ReactNode, MouseEvent, useCallback, useMemo } from "react";
 import _ from "lodash";
 import { createStyles, makeStyles, Theme } from "@material-ui/core/styles";
 import DetailsIcon from "@material-ui/icons/Details";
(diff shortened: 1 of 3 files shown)


src/data-table/ObjectsTable.tsx
```diff
@@ -78,35 +78,55 @@ export function ObjectsTable<T extends ReferenceObject = TableObject>(props: Obj
...
-                  <SearchBox
-                      value={searchValue}
-                      hintText={searchBoxLabel || "Search items"}
-                      onChange={handleSearchChange}
-                  />
-              </div>,
-              parentFilterComponents,
-          ]
-        : parentFilterComponents;
+    const filterComponents = useMemo(
```
tokland, line 89: That's the way I am also creating in-line memoized components, seems to work well.

tokland (approved):
Code ok, tested with project-monitoring-app, also ok. :+1: 

I've seen some problems with i18n, but that's for another PR:

https://github.com/EyeSeeTea/d2-ui-components/issues/141

xurxodev:
The blink seems to be fixed. However the tests continue to fail, so the cypress bug is confirmed.
cypress-bug

SferaDev:
> The blink seems to be fixed. However the tests continue to fail, so the cypress bug is confirmed.
> cypress-bug

@xurxodev Great, thanks for confirming!

We will need to track that cypress issue then...

PR #146, opened 2020-03-19
[Fix] Change typeof null returns object

SferaDev:
- https://github.com/EyeSeeTea/metadata-synchronization/issues/456


tokland (approved):
Interesting, I didn't know that. More info: https://stackoverflow.com/questions/18808226/why-is-typeof-null-object

PR #148, opened 2020-03-27
[DataTable] Allow overriding row styles from app

SferaDev:
New API: ``rowConfig?(row: T): RowConfig;``

RowConfig:

```
export interface RowConfig {
    style?: CSSProperties;
    disabled?: boolean;
    selectable?: boolean;
}
```

Usage example:

```
rowConfig={(_row: T) => ({
    style: { backgroundColor: "yellow"},
    disabled: true
})}
```

Behavior changes:

- Selectable: Disables the checkbox
- Disabled: Greys-out and disables pointer events (if selectable not set also disables the checkbox)
- Style: Overrides all pre-set styling in row (including changes applied by disabled/selectable)

### DIFF
--- a/src/data-table/DataTableBody.tsx
+++ b/src/data-table/DataTableBody.tsx
@@ -19,7 +19,7 @@ import {
     TableAction,
     TableColumn,
     TableSelection,
-    TableSorting
+    TableSorting,
 } from "./types";
 import { formatRowValue } from "./utils/formatting";
 import { isEventCtrlClick, parseActions, updateSelection } from "./utils/selection";
--- a/src/data-table/DataTableBody.tsx
+++ b/src/data-table/DataTableBody.tsx
@@ -11,8 +11,16 @@ import ChevronRightIcon from "@material-ui/icons/ChevronRight";
 import MoreVertIcon from "@material-ui/icons/MoreVert";
 import _ from "lodash";
 import React, { MouseEvent, useState } from "react";
-import { ReferenceObject, TableAction, TableColumn, TableSelection, TableSorting } from "./types";
-import { MouseActionsMapping, MouseActionMapping } from "./types";
+import {
+    MouseActionMapping,
+    MouseActionsMapping,
+    ReferenceObject,
+    RowConfig,
+    TableAction,
+    TableColumn,
+    TableSelection,
+    TableSorting
+} from "./types";
 import { formatRowValue } from "./utils/formatting";
 import { isEventCtrlClick, parseActions, updateSelection } from "./utils/selection";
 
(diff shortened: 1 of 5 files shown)


adrianq:
I guess this is ready to be merged, right @SferaDev? It seems to be used at https://github.com/EyeSeeTea/metadata-synchronization/pull/489

src/data-table/DataTableBody.tsx
```diff
@@ -38,17 +46,21 @@ const useStyles = makeStyles({
         background: "#E0E0E0",
     },
     disabledRow: {
-        backgroundColor: "#F5DFDF !important",
+        backgroundColor: "#EEEEEE",
```
adrianq, line 49: out of curiosity, why this change in the background color for disabled rows?

adrianq:
@tokland This seems to be working fine for metadata sync. Just wondering, if it has any impact on project monitoring. Can you have a look at it? If you haven't done it yet....

src/data-table/DataTableBody.tsx
tokland: @SferaDev  I guess now we should set the background color from the app `rowConfig`, now that we can? Although, maybe we could leave some (sane) different color, as default?

SferaDev:
I noticed that **Select all** includes disabled rows, I will add a fix for that to exclude them

src/data-table/DataTableBody.tsx
SferaDev: Exactly, I set ``grey`` color for disabled as it's the default in material and in the app we can now override the styling as we want.

src/data-table/DataTableBody.tsx
```diff
@@ -158,6 +171,7 @@ export function DataTableBody<T extends ReferenceObject>(props: DataTableBodyPro
                     onContextMenu={event => handleClick(event)}
                     role="checkbox"
                     selected={checked || indeterminate}
+                    style={style}
```
tokland, line 172: I am unable to make this work, I pass a style from the app, and it's indeed in this `style`, but it's not present on the `<tr>`. Am I missing something?

SferaDev: Not sure, I am able to do it on ``metadata-sync`` and I did not require the ``!important``, Perhaps ``project-monitoring`` has something in the global stylesheet?

tokland: It was me, I had to remove the "!important", cannot be used in an inline css. Working fine!

tokland (approved):
Code +:1+

https://github.com/EyeSeeTea/project-monitoring-app/pull/217

PR #149, opened 2020-05-01
[Fix] Properly style single urls in a bigger text

SferaDev:
- We styled the whole content if a single link was found
- Add ``react-linkify`` as we did in the past (see vaccination-app)

Before:

image

After:

image


adrianq (approved):
@tokland This works as expected. Will wait for your review before merging it

tokland (approved):
Tested with project-monitoring-app, all good :+1:

PR #152, opened 2020-05-12
[Fix] Add material-ui as a peer dependency

SferaDev:
### Implementation
- Add material-ui as peer dependency and move it to devDependencies so local compilation does not fail
- Fix some eslint errors

### Tested in:
  - Bulk-Load
  - metadata-sync

### DIFF
--- a/src/data-table/utils/selection.tsx
+++ b/src/data-table/utils/selection.tsx
@@ -70,11 +70,11 @@ export function getSelectionMessages<T extends ReferenceObject>(
             _(row)
                 .pick(childrenKeys)
                 .values()
-                .flatten()
                 .value()
         )
-        .flatten()
+        .flattenDeep()
         .value();
+
     const selection = _.differenceBy(tableSelection, childrenIds, "id");
 
     const allSelected = selection.length === pagination.total;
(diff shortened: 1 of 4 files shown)


src/data-table/utils/selection.tsx
```diff
@@ -70,11 +70,11 @@ export function getSelectionMessages<T extends ReferenceObject>(
             _(row)
                 .pick(childrenKeys)
                 .values()
-                .flatten()
                 .value()
         )
-        .flatten()
+        .flattenDeep()
```
tokland, line 75: [minor-subjective] When using `_.flattenDeep`, we lose a bit the perspective on the collection types we are working with. In this case, the old inner flatten + outer flatMap should be equivalent.

tokland (approved):
Cool, this way we reduce the bundle by having only one mui in node_modules and avoid problems with the injected styles.

src/data-table/utils/selection.tsx
adrianq: Guys, I am merging this PR as it is a dependency in this other [PR](https://github.com/EyeSeeTea/Bulk-Load/pull/75) that I have already merged. Feel free to create new PR if there is anything that worth to be changed!

PR #153, opened 2020-05-14
[OrgUnitsSelector] Fix selections from root

tokland:
On d2, the component used `list({root: ID, ...}`, but it turns out this property is not passed in the query string but in the URL: `/organisationUnits/$root?...`. 

https://docs.dhis2.org/2.33/en/developer/html/webapi_organisation_units.html#webapi_list_of_organisation_units#webapi_organisation_units_with_relations

We don't support yet custom models as organisationUnits, so I am using directly `d2.api.get(...)`. I'll open an issue on d2-api -> https://github.com/EyeSeeTea/d2-api/issues/42

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
         "type": "git",
         "url": "git+https://github.com/eyeseetea/d2-ui-components.git"
     },
-    "version": "1.3.2-beta.1",
+    "version": "1.3.3-beta.1",
     "main": "index.js",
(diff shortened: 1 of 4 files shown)


SferaDev (approved):
Tested on Bulk-Load and everything seems to work fine, thanks @tokland!

adrianq (approved):

PR #155, opened 2020-05-18
[OrgUnitTree] Allow filtering org unit tree

SferaDev:
### Implementation

- Add new property ``selectableIds`` to filter which organisations in the tree can be selected (either by checkbox or double click)

- Fix 414 error when passing thousands of roots to the tree

- Adapt Select All, Select by Group and Select by Level to be filtered by ``selectableIds`` if provided

### Others

- We should migrate these components (or at least ``OrgUnitsSelector`` to TS and functional components)

- I would add some changes to ``loading`` and ``error`` catching so that the UX is better when working with large OU trees (such as WHO's)

### DIFF
--- a/src/org-unit-select/common.js
+++ b/src/org-unit-select/common.js
@@ -18,18 +18,16 @@ const style = {
 style.button1 = Object.assign({}, style.button, { marginLeft: 0 });
 
 function addToSelection(orgUnits) {
-    const orgUnitArray = Array.isArray(orgUnits) ? orgUnits : orgUnits.toArray();
-    const addedOus = orgUnitArray.filter(ou => !this.props.selected.includes(ou.path));
-    const filteredOus = this.props.filter
-        ? addedOus.filter(({ id }) => this.props.filter.includes(id))
-        : addedOus;
+    const { selectableIds, selected } = this.props;
+    const additions = orgUnits.filter(({ id }) =>
+        selectableIds ? selectableIds.includes(id) : true
+    );
 
-    this.props.onUpdateSelection(this.props.selected.concat(filteredOus.map(ou => ou.path)));
+    this.props.onUpdateSelection(_.uniq([...selected, ...additions.map(ou => ou.path)]));
 }
 
 function removeFromSelection(orgUnits) {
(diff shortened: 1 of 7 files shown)


src/org-unit-select/common.js
```diff
@@ -18,15 +19,17 @@ const style = {
 style.button1 = Object.assign({}, style.button, { marginLeft: 0 });
 
 function addToSelection(orgUnits) {
-    const orgUnitArray = Array.isArray(orgUnits) ? orgUnits : orgUnits.toArray();
-    const addedOus = orgUnitArray.filter(ou => !this.props.selected.includes(ou.path));
+    const { selectableIds, selected } = this.props;
+    const additions = orgUnits.filter(({ id }) =>
+        selectableIds ? selectableIds.includes(id) : true
```
tokland: Equivalent, but it's perfectly fine as it is: `!selectableIds || selectableIds.includes(id)`

src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -141,17 +113,60 @@ export default class OrgUnitsSelector extends React.Component {
...
-        const response = api.models.organisationUnits.get(listOptions);
+        if (search) {
+            return api.models.organisationUnits.get({
+                ...baseOptions,
+                paging: true,
+                pageSize: 1000,
+                filter: { displayName: { ilike: search } },
+            });
+        } else if (rootIds) {
+            let cancel = false;
```
tokland, line 126: Keeping `cancel` in a closure, cool! This {getData, cancel} object is not exactly a full D2ApiResponse, but as it's all it's used in the code, is ok.

tokland (approved):
Cood looks good, tested in project-monitoring-app. I had only to add the typical `|| undefined` on one onClick to avoid a warning on the console: 98b932f

src/org-unit-select/common.js
SferaDev: Done!

src/org-units-selector/OrgUnitsSelector.jsx
SferaDev: Yeah, we should look at how to do this in the future and standardize it in all the apps.

adrianq:
Thanks guys!

PR #156, opened 2020-05-20
[ConfirmationDialog] Add additional help or info button

SferaDev:
- In Bulk-Load we will show a third button in the confirmation dialog.
- This is commonly known as ``help`` or ``info`` button and is placed at the bottom left corner

image

### DIFF
--- a/src/confirmation-dialog/ConfirmationDialog.tsx
+++ b/src/confirmation-dialog/ConfirmationDialog.tsx
@@ -5,6 +5,7 @@ import {
     DialogContent,
     DialogProps,
     DialogTitle,
+    makeStyles,
 } from "@material-ui/core";
 import _ from "lodash";
 import React, { ReactNode } from "react";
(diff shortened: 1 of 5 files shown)


src/confirmation-dialog/ConfirmationDialog.tsx
```diff
@@ -16,8 +17,10 @@ export interface ConfirmationDialogProps extends Partial<DialogProps> {
     description?: string | ReactNode;
     onSave?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
     onCancel?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+    onInfoAction?: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
     saveText?: string;
     cancelText?: string;
+    infoActionText?: string;
```
tokland, line 23: Generic note/comment for future components. Instead of pairs `onSomething`, `somethingText` we can have a unified object. Now that we have TS, the user can readily see what must be passed. For example: `something: {onClick: ..., text?: string}`. Drawback: we have to memoize it somehow, as it as an object.

tokland (approved):
Cool, just added the translations for new literal. Working on project-monitoring-app.

src/confirmation-dialog/ConfirmationDialog.tsx
SferaDev: Yeah, I guess here it would make sense to use ``useRef`` to store it and update it programmatically on each prop change but holding the ref intact so it doesn't change the object signature. It's probably the pattern I dislike the most of the react functional components...

tokland: Or `useMemo`. On this particular point, the original sin comes from JS: `{} === {} #=> false`

PR #157, opened 2020-05-21
[Proposal] Add new useConfirmationDialog hook

SferaDev:
Since we tend to have multiple dialogs in the same page, I'd like to propose a new API in d2-ui-components like this. I believe we already discussed this idea in the past, and I have a similar helper method in metadata-sync.

Example usage:

```ts
export function Component() {
    const [ModalComponent, updateModal, closeModal] = useModal(optionalInitialState);

    const onButtonClick = () => {
        updateModal({
            title: "Overwrite existing data values",
            description: "Are you sure?",
            onSave: () => performImport(info),
            onInfoAction: () => performImport(info, false),
            saveText: "Proceed",
            cancelText: "Cancel",
            infoActionText: "Import only new data values",
        });
    };

    return (
        <React.Fragment>
            <Button onClick={onButtonClick}>Import data</Button>
            <ModalComponent {...optionalSharedProps} />
        </React.Fragment>
    );
}
```

### DIFF
--- a/src/confirmation-dialog/useConfirmationDialog.tsx
+++ b/src/confirmation-dialog/useConfirmationDialog.tsx
@@ -2,10 +2,18 @@ import React, { useCallback, useState } from "react";
 import { ConfirmationDialog, ConfirmationDialogProps } from "./ConfirmationDialog";
 
 export interface ConfirmationDialogHookState extends Omit<ConfirmationDialogProps, "isOpen"> {
-    autoClose: boolean;
+    autoClose?: boolean;
 }
 
-export function useConfirmationDialog(initialState: ConfirmationDialogHookState | null = null) {
+export type ConfirmationDialogHookResult = [
+    (props: ConfirmationDialogProps) => JSX.Element | null,
+    (state: ConfirmationDialogHookState) => void,
+    () => void
+];
+
+export function useConfirmationDialog(
+    initialState: ConfirmationDialogHookState | null = null
+): ConfirmationDialogHookResult {
     const [modalState, updateModal] = useState<ConfirmationDialogHookState | null>(initialState);
     const closeModal = useCallback(() => updateModal(null), []);
 
--- a/src/confirmation-dialog/useConfirmationDialog.tsx
+++ b/src/confirmation-dialog/useConfirmationDialog.tsx
@@ -1,0 +1,43 @@
+import React, { useCallback, useState } from "react";
+import { ConfirmationDialog, ConfirmationDialogProps } from "./ConfirmationDialog";
+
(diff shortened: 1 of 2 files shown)


src/confirmation-dialog/useConfirmationDialog.tsx
```diff
@@ -0,0 +1,51 @@
...
+    () => void
+];
+
+export function useConfirmationDialog(
+    initialState: ConfirmationDialogHookState | null = null
+): ConfirmationDialogHookResult {
+    const [modalState, updateModal] = useState<ConfirmationDialogHookState | null>(initialState);
+    const closeModal = useCallback(() => updateModal(null), []);
+
+    const component = useCallback(
```
tokland, line 20: The first doubt that comes to mind: is it ok to return a component from a hook? They typically encapsulate behaviour, not view components. I found some discussions online, i.e: https://www.reddit.com/r/reactjs/comments/9yq1l8/how_do_you_feel_about_a_hook_returning_components/

The first obvious prevention people express is that the component identify may be new on each render. We have `useCallback`, which caches the component, but when `modalState` the component is created anew, right?

Not sure if possible, but ideally we'd have a normally defined component on the root level of the module that the hook uses. This component is very lean, so it's not really important here, just as a general concept.

src/confirmation-dialog/useConfirmationDialog.tsx
```diff
@@ -0,0 +1,51 @@
...
+                        if (onInfoAction) onInfoAction(event);
+                        if (autoClose) closeModal();
+                    }}
+                />
+            );
+        },
+        [closeModal, modalState]
+    );
+
+    return [component, updateModal, closeModal];
```
tokland, line 48: To be discussed: sometimes I think it's better to return an object unless they are crystal-clear tuples (ie. value/setValue in useState). This way it can grow (with optional fields) easily.

tokland:
So, in conclusion: It's mergeable. Only if you have some time, you can give it a try to the extracting the component thing.

tokland:
On the other hand, how about this split hook+component approach?  (if `useState` falls short, then a custom `useConfirmationDialogState` may be created):

```ts
export function Component() {
    const [props, setProps] = useState({ open: false, ... }); 

    const onButtonClick = useCallback(() => {
        setProps({
            open: true,
            title: "Overwrite existing data values",
            description: "Are you sure?",
            onSave: () => performImport(info),
            onInfoAction: () => performImport(info, false),
            saveText: "Proceed",
            cancelText: "Cancel",
            infoActionText: "Import only new data values",
        });
    }, [performImport]);

    return (
        <React.Fragment>
            <Button onClick={onButtonClick}>Import data</Button>
            <ConfirmationDialog {...props} />
        </React.Fragment>
    );
}
```

PR #159, opened 2020-05-26
Add sharing component

tokland:
- Adds Sharing react component
- SharingDialog has been removed, use a ConfirmationDialog as wrapper if needed.
- Usage example in README: https://github.com/EyeSeeTea/d2-ui-components/tree/feature/add-sharing-component#sharing

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,7 +308,7 @@ const MyWizard = props => {
 };
 ```
 
-# Sharings
+# Sharing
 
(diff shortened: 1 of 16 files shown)


SferaDev (approved):
LGTM, In the future we can review the following:

- Font aliasing when zoomed out screen
- Migrate rest of components to TS

PR #161, opened 2020-06-02
[OrgUnitsSelector] Add no option label

SferaDev:


adrianq:
@tokland I am merging this as I don't think it will cause any problem but have a look at it and let us know if you foresee any problem

PR #168, opened 2020-07-15
[Sharing] Add name as a valid title

SferaDev:
- In all the code we're always checking for ``displayName`` although that's true for most DHIS2 objects, in ``metadata-sync`` we send custom objects such as ``SyncRule`` that use simplified ``name`` instead of ``displayName/shortName/name``

- Also make some props that are optional, as optional types (public/external access)

- Allow customizing subtitle by props, Sharing component is also being used as CC notification list and "Who has access" makes no sense in that scenario

### DIFF
--- a/i18n/es.po
+++ b/i18n/es.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: i18next-conv\n"
-"POT-Creation-Date: 2020-07-01T08:06:53.907Z\n"
+"POT-Creation-Date: 2020-07-29T07:27:35.411Z\n"
 "PO-Revision-Date: 2019-02-14T11:21:26.165Z\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
@@ -163,10 +163,10 @@ msgstr ""
 msgid "Can capture and view"
 msgstr ""
 
-msgid "Created by"
+msgid "Who has access"
 msgstr ""
 
-msgid "Who has access"
+msgid "Created by"
 msgstr ""
 
 msgid "Add users and user groups"
--- a/i18n/es.po
+++ b/i18n/es.po
@@ -134,46 +134,46 @@ msgid "Remove"
 msgstr "Quitar"
 
 msgid "This access cannot be removed"
-msgstr ""
+msgstr "No se puede eleminar este acceso"
 
 msgid "External access"
-msgstr ""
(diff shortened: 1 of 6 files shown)


adrianq:
@tokland there is no problem with this PR in MDSync. I will merge it as soon as you let me know you are happy with it

i18n/es.po
```diff
@@ -163,10 +163,10 @@ msgstr ""
 msgid "Can capture and view"
 msgstr ""
 
-msgid "Created by"
+msgid "Who has access"
```
tokland, line 166: I'll complete the ES translation, now that we are in it.

src/sharing/types.ts
```diff
@@ -11,12 +11,13 @@ export interface SharingRule {
 
 export interface SharedObject {
     id: Id;
+    name: string;
```
tokland: `name` as a required prop? since the expression using it is `displayName || name`, I guess you wanted it to make optional, right?

src/sharing/Table.tsx
```diff
@@ -110,10 +111,11 @@ class Table extends React.Component<TablePropsWithStyles> {
 
     render() {
         const { allowPublicAccess, allowExternalAccess } = this.props.meta.meta;
```
tokland: Now that they are optional, we may define default values. `publicAccess=true` and `externalAccess=false`?

src/sharing/types.ts
```diff
@@ -25,8 +26,8 @@ export type ShareUpdate = Partial<
 
 export interface MetaObject {
     meta: {
```
tokland: Would you make `meta` itself optional? now you're forced to send at least an empty object.

tokland (commented):
Added some i18n literals. Some minor comments:

src/sharing/types.ts
SferaDev: Actually it makes sense to me that ``displayName`` isthe optional one and if defined used and otherwise use the required ``name``

src/sharing/Table.tsx
SferaDev: Sure thing

src/sharing/types.ts
SferaDev: Yes looks good to me

SferaDev:
LGTM

tokland (approved):
Changes proposed, commited. Published: 1.6.1-beta.5

PR #170, opened 2020-07-17
Feature/allow no primary action

tokland:
Required by https://github.com/EyeSeeTea/project-monitoring-app/pull/261

- Sometimes we don't want a default action to execute as the primary action, this PRs removes the OR fallback.
- Also: when rendering custom components in the table, `handleClick` was getting events with targets that have no tagName, which resulted in an error. Add a check before using it.

### DIFF
--- a/src/data-table/DataTableBody.tsx
+++ b/src/data-table/DataTableBody.tsx
@@ -123,6 +123,7 @@ export function DataTableBody<T extends ReferenceObject>(props: DataTableBodyPro
 
         const handleClick = (event: MouseEvent<unknown>) => {
             const { tagName, type = null } = event.target as HTMLAnchorElement;
+            if (!tagName) return;
             const isCheckboxClick = tagName.localeCompare("input") && type === "checkbox";
             const activeSelection = _.reject(selected, { indeterminate: true });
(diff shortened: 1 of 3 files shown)


src/data-table/DataTableBody.tsx
```diff
@@ -94,7 +94,7 @@ export function DataTableBody<T extends ReferenceObject>(props: DataTableBodyPro
     function createRow(row: T, index: number | string, level = 0) {
         const labelId = `data-table-row-${index}`;
         const defaultAction = parseActions([row], availableActions)[0];
```
SferaDev: Not sure if we actually need ``defaultAction`` for anything right now. Its only usage is to show/hide the three button icon.

```suggestion
        const showRowActions = parseActions([row], availableActions).length > 0;
```

And in line 214/215

```
                        {showRowActions && (
                            <Tooltip title={i18n.t("Actions")}>
                                <IconButton onClick={event => contextualAction(event)}>
                                    <MoreVertIcon />
                                </IconButton>
                            </Tooltip>
                        )}
```

SferaDev (approved):

PR #171, opened 2020-07-30
[Cleanup] Remove Old objects table and node-sass

### LINKED ISSUE #115: Remove old objects-table
This way we can remove the heavy and burdensome node-sass library.

We can keep a branch with the old objects table (should we ever need an update on old projects using it), and publish in a different channel.

SferaDev:
### References

- * **Issue:** Closes #115 

### Implementation

- This is a breaking change and probably should be part of a `2.0.0` release
- In theory only ``vaccination-app`` should be affected by this change, and if any non-EyeSeeTea usages


tokland (approved):
Cool, glad to see node-sass go away! I just removed some commented literals in i18n files.

adrianq:
Thanks, guys! I just bumped the version to 2.0.0 as you suggested, released a beta version and updated MDSync app

PR #173, opened 2020-08-21
[ObjectsTable] Refine parsing of instance name for date-like names

tokland:
Required by https://github.com/EyeSeeTea/metadata-synchronization/pull/597

We are using ISO 8601 ([link](https://www.w3.org/TR/NOTE-datetime)) to determine if a string looks like a date, so `YYYY` is a valid date. I've now required at least something looking like "YYYY-MM-DD" to consider it a date, which seems an acceptable compromise.

### DIFF
--- a/src/data-table/utils/formatting.tsx
+++ b/src/data-table/utils/formatting.tsx
@@ -20,7 +20,7 @@ function defaultFormatter(value: any): ReactNode {
         );
     } else if (value.displayName || value.name || value.id) {
         return value.displayName || value.name || value.id;
-    } else if (moment.isDate(value) || moment.isMoment(value) || isStringDate(value)) {
+    } else if (isValidDate(value)) {
         return moment(value).format("YYYY-MM-DD HH:mm:ss");
     } else {
         return <Linkify key={value}>{value}</Linkify>;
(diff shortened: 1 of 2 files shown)


src/data-table/utils/formatting.tsx
```diff
@@ -40,3 +36,12 @@ export function formatRowValue<T extends ReferenceObject>(
 }
 
 export const defaultRowConfig = () => ({} as RowConfig);
+
+function isStringDate(value: any): boolean {
+    // Avoid dubious positives by skipping strings that do not contain at least year, month and day.
+    if (typeof value === "string" && value.length >= "YYYY-MM-DD".length) {
+        return moment(value, moment.ISO_8601, true).isValid();
```
SferaDev: Not completely sure about this check.

Would it make sense to check for multiple valid formatting in strict mode instead of ISO_8601:

``formats = [DD/MM/YYYY, DD-MM-YYYY, MM/DD/YYYY, MM-DD-YYYY]``

And use ``moment(value, format, true).isValid()``?

SferaDev: As discussed earlier this week, adding the check from ``creationData()``

SferaDev (approved):

PR #187, opened 2020-12-14
Feature/orgunit selector select by program and select intersection

xurxodev:
Required by https://github.com/EyeSeeTea/metadata-synchronization/pull/724

- [x] Add dropdown in the org unit selector to select by program
- [x] Select by program hide by default
- [x] Add select intersection button

**Note**: The org unit selector is an old code in javascript and I think with duplicate code, now I have made the snowball a little bigger. In the future, if we migrate to typescript I think will be necessary to rethink the related components (select by level, org unit group, program, and all selects) to create a unique function to retrieve data (like a repository) with multiple params instead of to realize similar requests to dhis2 in every component.


adrianq:
@tokland tested from MDSync. It works fine. Just updated some filter names and release a version to the beta channel. Will wait for you to confirm everything is fine in particular for SP

src/org-unit-select/OrgUnitSelectAll.component.js
```diff
@@ -65,6 +69,101 @@ class OrgUnitSelectAll extends React.Component {
         }
     }
 
+    getRelativeLevelFilter(api, level, currentRoot) {
+        const rootLevel =
+            currentRoot.level || currentRoot.path
+                ? this.props.currentRoot.path.match(/\//g).length
```
tokland, line 77: prop `currentRoot` is not set as required, so this may fail (not here, as the parent has a previous if, but in the general case).

src/org-unit-select/OrgUnitSelectAll.component.js
```diff
@@ -65,6 +69,101 @@ class OrgUnitSelectAll extends React.Component {
...
+                : NaN;
+        return level - rootLevel;
+    }
+
+    getOrgUnitsByFilters() {
+        const { api } = this.context;
+
+        const cacheKey = `${this.props.selectedFilters.level}-${this.props.selectedFilters.orgUnitGroupId}-${this.props.selectedFilters.programId}`;
```
tokland: [minor, subjective] De-structuring the common object (this.props) may lighten the method. Maybe use join("-") to build the key?

src/org-unit-select/OrgUnitSelectAll.component.js
```diff
@@ -65,6 +69,101 @@ class OrgUnitSelectAll extends React.Component {
...
+
+        const level = this.props.selectedFilters.level
+            ? this.props.selectedFilters.level
+            : undefined;
```
tokland: `const x = y ? y : undefined` -> `const x = y || undefined`. 

But is that required to capture non-undefined falsy values or what's the idea?

src/org-unit-select/OrgUnitSelectAll.component.js
```diff
@@ -124,6 +240,9 @@ OrgUnitSelectAll.propTypes = {
     // If currentRoot is set, only org units that are descendants of the
     // current root org unit will be added to or removed from the selection
     currentRoot: PropTypes.object,
+
+    // selected filters values: level, orgUnitGroupId, programId
+    selectedFilters: PropTypes.object,
```
tokland: Ideally, we'd define the prop with `PropTypes.shape`. That will be also very useful the day we port this to Typescript.

src/org-unit-select/OrgUnitSelectByProgram.component.js
```diff
@@ -0,0 +1,132 @@
...
+        this.handleChangeSelection = handleChangeSelection.bind(this);
+
+        this.getOrgUnitsForProgram = this.getOrgUnitsForProgram.bind(this);
+        this.handleSelect = this.handleSelect.bind(this);
+        this.handleDeselect = this.handleDeselect.bind(this);
+    }
+
+    getOrgUnitsForProgram(programId, ignoreCache = false) {
+        const { api } = this.context;
+        return new Promise(resolve => {
```
tokland, line 34: Not that it's wrong, but I guess we can just return `Promise.resolve(value)` in the non-async branches.

src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -299,20 +324,61 @@ export default class OrgUnitsSelector extends React.Component {
                                                         onUpdateSelection={
                                                             this.handleSelectionUpdate
                                                         }
+                                                        onItemSelection={selectedItem => {
```
tokland: As a general rule, avoid arrow function as props, it can be moved to a method.

src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -299,20 +324,61 @@ export default class OrgUnitsSelector extends React.Component {
...
+                                                            this.setState({
+                                                                ...this.state,
```
tokland: `this.setState` merges the previous state by default, this `...this.state` should not be necessary.

src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -299,20 +324,61 @@ export default class OrgUnitsSelector extends React.Component {
...
+                                                                selectedFilters: {
+                                                                    ...this.state.selectedFilters,
```
tokland: As a rule of thumb, setState calls should, when updating, use the form `this.setState(oldState => ....)`, instead of using `this.state`, which may be stale.

tokland (approved):
I had no problems with project-monitoring, and used the new program selector when I enabled the flag. 

And the code looks very good. Just some minor comments about the code, which are not stoppers if we need a quick merge:

adrianq:
There is no rush in merging this PR so feel free to discuss and change these comments. Just remember to release a beta version once you are done so we can validate it in MDSync and SP

src/org-unit-select/OrgUnitSelectAll.component.js
xurxodev: current root must be non required, I have modified this method

src/org-unit-select/OrgUnitSelectAll.component.js
xurxodev: Fixed!

src/org-unit-select/OrgUnitSelectAll.component.js
xurxodev: Fixed!

src/org-unit-select/OrgUnitSelectAll.component.js
xurxodev: Fixed!

src/org-unit-select/OrgUnitSelectByProgram.component.js
xurxodev: I don't understand you, the code is equal to other methods in existed components as getOrgUnitsForGroup in OrgUnitSelectByGroup.component

src/org-units-selector/OrgUnitsSelector.jsx
xurxodev: Fixed!

src/org-units-selector/OrgUnitsSelector.jsx
xurxodev: Fixed!

src/org-units-selector/OrgUnitsSelector.jsx
xurxodev: Fixed!

PR #189, opened 2020-12-15
Allow link to app

### LINKED ISSUE #188: Support link with apps on DEV
Currently, you get an error "more than one copy of React in the same app" when trying to "yarn link". Probably, react/material should be only in peerDependencies.

tokland:
Closes #188

That's a bit tricky. The key is to move dependencies to peerDependencies so it uses React from the app. The problem is that now we don't have React on this repo, so, for example, we cannot run tests. I've disabled them for now (only a DatePicker example test is present).

Also, when testing, sometimes you get the "2 reacts loaded" error, when it should not. Delete node_modules and start again. Removing react/react-dom from package.json dependencies (keep only as peer) should work.

One solution would be to generate a build/package.json on-the-fly, without react and react-dom devDepedencies. But I've not found anyone doing something like this, so it probably needs a little more investigation.

Let's discuss it in some team meeting.


SferaDev:
We can address this tomorrow!

To my understanding anything that requires an actual version of React would be under a CRA. We already have the examples/demo-app and the tests should live there. 

In some libraries they follow the following structure:

package.json: Hosts scripts only
- lib:
  node_modules
  package.json: Hosts library deps + react as a peer dep
- demo:
  node_modules
  package.json: Hosts real deps

tokland:
- [ ] `example/` create-react-app (with tests)
- [ ] main: only build.

tokland:
We can merge this and create an issue with the tasks in the comment.

PR #190, opened 2020-12-17
Fix bug with progress bar hidden by default

xurxodev:
Required by https://github.com/EyeSeeTea/metadata-synchronization/pull/759

- [x] Fix bug when the indeterminate progress bar is hidden by default.


tokland (approved):
Now we see the spinner in project-monitoring-app, thx!

PR #195, opened 2020-12-23
[Fix] Move material-ui pickers to dependencies

### LINKED ISSUE #194: material-ui/pickers dependencies
We should add package to `dependencies` or `peerDependencies`. @SferaDev, what's your take?

SferaDev:
- Issue: Closes #194


tokland (approved):
Tested with SP/MdSync, in normal and link mode, it works well, so let's go with the dependency. Added a couple minor commits about the build.

PR #196, opened 2021-01-05
[Feature] Add dropdown components

SferaDev:
We are using Dropdowns in almost every app with different implementations. We are adding it once again form ``predictors-extended`` so I'm bringing the code from ``project-monitoring`` with small changes introduced in ``metadata-sync``.

  - Added type narrowing for unions
  - Added className overrides for styled-components
  - Migrated to NamedExports instead of DefaultExports﻿

### DIFF
--- a/src/dropdown/Dropdown.tsx
+++ b/src/dropdown/Dropdown.tsx
@@ -1,0 +1,57 @@
+import { MenuItem, Select } from "@material-ui/core";
+import React from "react";
+import i18n from "../utils/i18n";
+import { DropdownForm, DropdownItem } from "./GenericDropdown";
+
+export interface DropdownProps<Value extends string = string> {
+    className?: string;
+    items: DropdownItem[];
+    onChange: (value: Value | undefined) => void;
+    label?: string;
+    value?: Value;
+    hideEmpty?: boolean;
+}
+
+interface SelectProps {
+    className?: string;
+    label?: string;
+}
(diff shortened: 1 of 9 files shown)


src/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,57 @@
...
+    return (
+        <SelectWrapper className={className} label={label}>
+            <Select
+                data-cy={label}
+                value={selectValue}
+                onChange={ev => onChange((ev.target.value as string) || undefined)}
+                MenuProps={{
+                    getContentAnchorEl: null,
+                    anchorOrigin: { vertical: "bottom", horizontal: "left" },
+                }}
```
tokland, line 46: Can we make onChange and MenuProps immutable?

tokland (approved):
Test with sp-project, all good. Just a minor thing about a couple of props, if you feel it's necessary.

src/dropdown/Dropdown.tsx
SferaDev: As this is really minor, leaving it for now.

We might revisit them as soon as the component is used on other projects.

Maybe we could migrate the multiple dropdown on SP and revisit this minor things and maybe other features. See https://github.com/EyeSeeTea/project-monitoring-app/blob/development/src/components/dropdown/MultipleDropdown.tsx

PR #198, opened 2021-01-20
[Feature] Add ObjectsList component

SferaDev:


tokland (approved):
A minor fix on an import and some minor refactors to immutabilize props. Package published. All good!

PR #204, opened 2021-03-29
[Fix] Children row selection

SferaDev:
- Fix executing contextual actions on children rows of level >= 3


xurxodev (approved):
Thanks @SferaDev, I have tested the fix in MD sync a works

PR #208, opened 2021-04-09
[DataTable] Allow custom selection messages

tokland:
Required by https://app.clickup.com/t/gv514v

I checked if there was a direct override mechanism for i18n files from the main app, but it does not seem to be supported. So let's just use an optional (partial) prop with the translations.

### DIFF
--- a/src/utils/i18n.ts
+++ b/src/utils/i18n.ts
@@ -1,4 +1,3 @@
-// @ts-ignore
 import i18n from "@dhis2/d2-i18n";
 
 function t(key: string, options?: object): string {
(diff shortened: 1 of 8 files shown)


src/utils/i18n.ts
```diff
@@ -1,17 +1,21 @@
+// @ts-ignore
```
SferaDev: In a ``modules.d.ts``:

```ts
declare module "@dhis2/d2-i18n" {
    export function t(value: string): string;
    export function t(value: string, options?: {
        nsSeparator?: boolean;
        [key: string]: any;
    }): string;
}
```

tokland: Done. Thanks @SferaDev, I was kind of lazy :)

PR #217, opened 2021-11-25
ColumnSelectorDialog: allow column ordering

oventi:
### :pushpin: References
-   **Issue:** Closes https://app.clickup.com/t/1kxx8pg

### :memo: Implementation
- Support options to pick visible/not visible columns only and more complex column visibility with ordering (via the Transfer component)
- By default, use Transfer, unless reorder property is false
- Some type checking is bypassed as there is a conflict between a `string[]` and a `(keyof T)[]` types
- This modification was requested for `d2-reports`. No code modification is needed
- Any other code that uses ColumnSelectorDialog should be tested to ensure the column ordering does not break anything

### DIFF
--- a/src/data-table/ColumnSelectorDialog.tsx
+++ b/src/data-table/ColumnSelectorDialog.tsx
@@ -16,8 +16,7 @@ interface ColumnSelectorDialogProps<T extends ReferenceObject> {
 export function ColumnSelectorDialog<T extends ReferenceObject>(
     props: ColumnSelectorDialogProps<T>
 ) {
-    const reorder = typeof props.reorder === "undefined" ? true : props.reorder;
-    const { columns, visibleColumns, onChange, onCancel } = props;
+    const { columns, visibleColumns, onChange, onCancel, reorder = true } = props;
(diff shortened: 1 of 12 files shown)


src/data-table/ColumnSelectorDialog.tsx
```diff
@@ -1,31 +1,28 @@
...
+    reorder?: boolean;
     onChange: (visibleColumns: (keyof T)[]) => void;
     onCancel: () => void;
 }
 
 export function ColumnSelectorDialog<T extends ReferenceObject>(
     props: ColumnSelectorDialogProps<T>
 ) {
+    const reorder = typeof props.reorder === "undefined" ? true : props.reorder;
     const { columns, visibleColumns, onChange, onCancel } = props;
```
SferaDev:
```suggestion
    const { columns, visibleColumns, onChange, onCancel, reorder = true } = props;
```

SferaDev (approved):

SferaDev:
You have the code available on npm: https://www.npmjs.com/package/@eyeseetea/d2-ui-components/v/2.6.9-beta.4

PR #218, opened 2021-11-26
[Feature] All ids on objects list table

SferaDev:


src/data-table/utils/selection.tsx
```diff
@@ -105,7 +105,7 @@ export function getSelectionMessages<T extends ReferenceObject>(
     const selectionInOtherPages = _.differenceBy(selection, rows, "id");
     const allSelectedInPage = _.differenceBy(rows, selection, "id").length === 0;
     const multiplePagesAvailable = total > rows.length;
-    const selectAllImplemented = ids.length === total;
+    const selectAllImplemented = ids.length > 0;
```
SferaDev, line 108: Somethings the pager is incorrect (reports more users than rows, this fixes the problem by enabling select all if you provide the prop required for the feature to work

tokland (approved):
LGTM

PR #223, opened 2022-01-19
[Fix] Initial state on table list

SferaDev:
- [Fix] Initial sorting for a column that is not the first was not set (leading to change re-ordering on selection)
- [Feature] Allow setting initial selection


tokland (approved):
Code looks good and worked just fine with project-monitoring-app :+1:

PR #227, opened 2022-07-08
Add switch to OrgUnitsSelector to use orgUnit.shortName

jordibc:
Reference ticket: https://app.clickup.com/t/2ghev8r

We want the OU tree to be able to show shortnames instead of just the
default displaynames. This commit adds a Switch inside a FormControlLabel
between the search box and the OU tree of the OrgUnitsSelector
component to do just that.

The trickiest part was adding:
  key={useShortNames}
to the OrgUnitTree component, because it not having a unique key before
meant that react never knew it had to update it after changing the
state of the OrgUnitsSelector component.

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2022-07-07T22:41:42.386Z\n"
-"PO-Revision-Date: 2022-07-07T22:41:42.386Z\n"
+"POT-Creation-Date: 2022-07-11T04:52:08.113Z\n"
+"PO-Revision-Date: 2022-07-11T04:52:08.113Z\n"
 
 msgid "Cancel"
 msgstr ""
@@ -127,6 +127,9 @@ msgstr ""
 msgid "For all organisation units"
 msgstr ""
 
+msgid "Use short names"
+msgstr ""
+
 msgid "Remove"
 msgstr ""
 
@@ -179,7 +182,4 @@ msgid "Previous"
 msgstr ""
 
 msgid "Next"
-msgstr ""
-
-msgid "Use short names"
 msgstr ""
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2021-11-24T07:20:30.326Z\n"
-"PO-Revision-Date: 2021-11-24T07:20:30.326Z\n"
+"POT-Creation-Date: 2022-07-07T22:41:42.386Z\n"
+"PO-Revision-Date: 2022-07-07T22:41:42.386Z\n"
 
 msgid "Cancel"
 msgstr ""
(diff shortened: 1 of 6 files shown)


jordibc:
This is how it looks now, deactivated:

Screenshot from 2022-07-08 10-03-21

And activated (note that the shortname is the same for these org units, except the root one, which doesn't have one):

Screenshot from 2022-07-08 10-03-35

src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -1,4 +1,4 @@
-import { Card, CardContent } from "@material-ui/core";
```
tokland, line 1: This file is not prettified (`yarn prettify`). Having vscode with extension `Prettier` + setting "Format on save" enabled should solve the problem altogether.

i18n/en.pot
```diff
@@ -180,3 +180,6 @@ msgstr ""
 
 msgid "Next"
 msgstr ""
+
+msgid "Use short names"
+msgstr ""
```
tokland: Running `yarn localize` moves these new translation in the file, does it happen also for you?

src/org-unit-tree/OrgUnitTree.component.js
```diff
@@ -50,6 +50,7 @@ class OrgUnitTree extends React.Component {
         super(props);
 
         this.state = {
+            useShortNames: props.useShortNames,
```
tokland: `useShortNames` is not updated within the component, right? then it can be simply used from the props.

src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -312,7 +321,20 @@ export default class OrgUnitsSelector extends React.Component {
                         <div style={styles.contentItem}>
                             {roots.map(root => (
                                 <div key={root.path} className={`ou-root ${getClass(root)}`}>
+                                    <FormControlLabel
```
tokland: This new switch will now always be visible to users. However, in most cases, we will want to keep it as-is, a lot of projects use this component. Proposal: Use two props for `OrgUnitTree` so the caller has more control:

- `showShortName?: boolean` (defaults to false)
- `showNameSetting?: boolean` (defaults to false)

src/org-units-selector/OrgUnitsSelector.jsx
jordibc: Right, and set it as the default formatter in the general preferences. I changed my programming setup and missed it, even though you had already showed me this. Fixed now. Thanks!

i18n/en.pot
jordibc: It does happen to me. I wasn't that familiar with the proper use of localize, so I missed it. I understand now that one should run it too when adding new texts. Fixed now, thanks!

src/org-unit-tree/OrgUnitTree.component.js
jordibc: Totally. Done in https://github.com/EyeSeeTea/d2-ui-components/commit/62085fc06ed

i18n/en.pot
jordibc: Ref: https://github.com/EyeSeeTea/d2-ui-components/commit/83b3c36285

src/org-units-selector/OrgUnitsSelector.jsx
jordibc: Makes total sense. Fixed in https://github.com/EyeSeeTea/d2-ui-components/commit/97676e17

jordibc:
Thanks @tokland , I found all the points valid and I think they are all addressed with the last commits, so I'm sending a new request for review.

jordibc:
Just to let you know, I tried to use the component from user-extended-app, with:

```html
<OrgUnitsSelector
    ...
    showShortName={xxx}
    showNameSetting={yyy}
/>
```

for all true/false combinations of `xxx` and `yyy` and it seems to work.

tokland (approved):
Cool, all good by me!

ifoche (approved):
thanks @jordibc  @tokland

PR #228, opened 2023-04-28
Removing sort icon if all columns are 

eperedo:
### :pushpin: References

-   **Issue:** Closes [https://app.clickup.com/t/865c777eq](https://app.clickup.com/t/865c777eq)

### :memo: Implementation

Current implementation was not checking if the column has the sortable property as true. Also I added the **hideSortIcon** to hide the icon if the column is not active. Right now is still in the dom with opacity 0.

https://v4.mui.com/api/table-sort-label/#props

### :art: Screenshots

All columns with sortable = false
table_no_sort

All columns with sortable = true
table_sort

Dom without disabling the hideSortIcon prop
table_icon_opacity_0

Dom disabling the hideSortIcon if column is not sortable
table_sortable_false_hide_icon

### :fire: Testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
         "type": "git",
         "url": "git+https://github.com/eyeseetea/d2-ui-components.git"
     },
-    "version": "2.6.12-beta.4",
+    "version": "2.6.12-beta.6",
     "main": "index.js",
     "types": "index.d.ts",
     "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
         "type": "git",
         "url": "git+https://github.com/eyeseetea/d2-ui-components.git"
     },
-    "version": "2.6.12-beta.3",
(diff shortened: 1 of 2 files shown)


adrianq:
@eperedo I haven't looked into this PR yet. I will wait for @tokland review first but please remember to update CPR to include this fix. As we discussed, we may need to validate the update of d2-ui-component does not break anything in the layout. You can do the update in your current branch feat/upg-modal-selector Both @tokland and I can publish a beta version of d2-ui-components in npm just give any of us a shout

tokland:
The MUI layout is mangled on development. When finished, to check that all works as expected (specially when one has changed styles - not here) when the npm package is not yet uploaded, we can run:

```
$ yarn add @eyeseetea/d2-ui-components@/path/to/d2-ui-components/build 
```

tokland (approved):
Just changed to the latest free beta version

PR #233, opened 2023-11-08
notify if use short names switch change

eperedo:
### :pushpin: References

* **Issue:** Closes https://app.clickup.com/t/860ryt00g

### :memo: Implementation

- Adding function prop **onUseShortNamesChange** so we can know the current value of the **Use short names** switch.

This is for bulk load because we need to know the current value of the switch in order to generate the org units metadata using the **displayName** or **displayShortName** property in templates.

### :fire: Notes for the reviewer

@ifoche can you help me publishing a new beta version for bulk load, please?

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
         "type": "git",
         "url": "git+https://github.com/eyeseetea/d2-ui-components.git"
     },
-    "version": "2.7.0_beta.2",
+    "version": "2.7.0-beta.3",
     "main": "index.js",
     "types": "index.d.ts",
     "peerDependencies": {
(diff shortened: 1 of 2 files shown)


ifoche (commented):
@eperedo you mention that you need a beta in d2-ui-components but... Can't you use the same one from the previous PR? Is there anything new you're adding here to d2-ui-components?

eperedo:
> @eperedo you mention that you need a beta in d2-ui-components but... Can't you use the same one from the previous PR? Is there anything new you're adding here to d2-ui-components?

Yes, right now the org. unit component manage the state internally so from Bulk Load (or any other app) we actually cannot know what's the state of the "Use short names" switch. This PR include a prop to the component so we can get the current value of the switch.

adrianq:
@eperedo I was around so I have just published a beta for you: v2.7.0-beta.3

ifoche (approved):
thanks @eperedo

PR #235, opened 2024-01-29
add completed prop for wizard

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8693muw6j

### :memo: Implementation

In the data quality app we need a way to indicate that a step is already completed. The `<Step />` already has support for this, but right now the value is hardcoded to `false`. I'm adding a new prop to give users the control of this behavior.

This is how it looks when the prop `completed` is true:

image

**Wizard with step "Summary" mark as completed**
image

@adrianq @ifoche can you help publishing a new beta version for this please?

### DIFF
--- a/src/wizard/Stepper.tsx
+++ b/src/wizard/Stepper.tsx
@@ -27,7 +27,7 @@ export const Stepper: React.FC<StepperProps> = ({
             {steps.map((step, index) => (
                 <Step
                     key={step.key}
-                    completed={false}
+                    completed={step.completed || false}
                     disabled={index > lastClickableStepIndex}
                     className={"Wizard-Step"}
                 >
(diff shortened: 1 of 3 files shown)


src/wizard/Stepper.tsx
```diff
@@ -27,7 +27,7 @@ export const Stepper: React.FC<StepperProps> = ({
             {steps.map((step, index) => (
                 <Step
                     key={step.key}
-                    completed={false}
+                    completed={step.completed || false}
```
tokland, line 30: It's 100% ok, just as a note, other options for boolean defaults: `Boolean(step.completed)`, `step.completed ?? false`.

tokland (approved):
All good. Tested with project-monitoring-app.

PR #238, opened 2024-03-11
Feat: clickable stepper

mariaozamiz:
### :pushpin: References

-   **Issue:** [DQ: Make Stepper clickable](https://app.clickup.com/t/86940n8cx)

### :memo: Implementation

Installed Storybook to easily instance and test components.
Add an optional prop called stepsDisableUntilReached, true by default. 

Captura de pantalla 2024-03-11 a las 11 56 58

yarn storybook to check isolated Wizard component


eperedo (commented):
@mariaozamiz Maybe I'm missing something, but I think we already have support for this functionality using the `lastClickableStepIndex` prop.

```tsx
<Wizard
  // this is basically telling the component
  // we can click until the last step
  lastClickableStepIndex={steps.length}
  steps={steps}
/>
```
Demo
[Data-Quality_click_steps.webm](https://github.com/EyeSeeTea/d2-ui-components/assets/461124/888bc80b-dc20-41f2-9f17-7b43032111c7)

mariaozamiz:
You're right! I missed that one. Nevertheless, we should handle the disabled state of the buttons too, otherwise, we're sending visual cues of "deactivation". Maybe we can connect both props?

mariaozamiz:
The change is no longer relevant as we are addressing the stepper in a different way now

PR #239, opened 2024-03-28
filter org. units by selectableLevels

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/86945xe2m

### :memo: Implementation

Right now if you're using the `OrgUnitsSelector` component with the `selectableLevels` prop to limit the org units.

```tsx
<OrgUnitsSelector
    selectableLevels={[1, 2]}
 />
```

it'll render this:

image

which is correct, but if you click on the `SELECT ALL` button it won't apply the `selectableLevels` value to filter the selected org units:

image

which could be confusing because initially you only want to work with org. units in levels 1 and 2. Same case for the `SELECT` button in the Organisation Group selector.

This PR change the behavior so the `selectableLevels` is being considered in the selected org units.

[UPDATE 2024-04-01]

I'm including another feature for the Wizard component in order to have custom icons for each step.

[UPDATE 2024-05-02]
Adding props to add custom css and sticky headers to the `DataTable` component

image
I'm reusing the same PR because I need both changes in the same app, but let me know if I should create another PR.

**Selector with `selectableLevels={[1, 2]}` (current behavior)**
[Data-Quality_selectable_orgunits_before_fix.webm](https://github.com/EyeSeeTea/d2-ui-components/assets/461124/e531c28e-4d0e-46a5-ac2f-907f952a7022)

**Selector with `selectableLevels={[1, 2]}` (new behavior)**
[Data-Quality_selectable_orgunits.webm](https://github.com/EyeSeeTea/d2-ui-components/assets/461124/d6862dd1-2fcd-4f6d-a6fe-b8817532f36d)

**Custom icons for each step in wizard**
image

#86945xe2m

### DIFF
--- a/src/org-unit-select/OrgUnitSelectAll.component.js
+++ b/src/org-unit-select/OrgUnitSelectAll.component.js
@@ -52,8 +52,13 @@ class OrgUnitSelectAll extends React.Component {
         } else {
             this.setState({ loading: true });
 
+            const filters =
+                this.props.selectableLevels && this.props.selectableLevels.length > 0
+                    ? { level: { in: this.props.selectableLevels } }
+                    : undefined;
+
             this.context.api.models.organisationUnits
-                .get({ fields: { id: true, path: true }, paging: false })
+                .get({ fields: { id: true, path: true }, paging: false, filter: filters })
                 .getData()
                 .then(({ objects }) => {
                     this.addToSelection(objects);
--- a/src/org-unit-select/OrgUnitSelectAll.component.js
+++ b/src/org-unit-select/OrgUnitSelectAll.component.js
@@ -5,6 +5,7 @@ import Button from "@material-ui/core/Button";
 import i18n from "../utils/i18n";
 
 import { addToSelection, removeFromSelection } from "./common";
+import { isSelectableLevelsDefined } from "./OrgUnitSelectByGroup.component";
 
 const style = {
     button: {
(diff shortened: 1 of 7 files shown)


ifoche:
Task linked: [CU-86945xe2m DQ: Fix Country selector](https://app.clickup.com/t/86945xe2m)

eperedo:
@adrianq @tokland can you please help me with a beta version?

tokland:
@eperedo: Done: @eyeseetea/d2-ui-components@2.9.0-beta.1. You can set the beta yourself (package.json->version)depending on the change being a release/patch.

Before the final approval of this PR, make sure you've merged development into this branch.

src/org-unit-select/OrgUnitSelectAll.component.js
```diff
@@ -52,8 +52,13 @@ class OrgUnitSelectAll extends React.Component {
         } else {
             this.setState({ loading: true });
 
+            const filters =
+                this.props.selectableLevels && this.props.selectableLevels.length > 0
```
tokland: This is `isSelectableLevelsDefined`, right?

src/org-unit-select/OrgUnitSelectByGroup.component.js
```diff
@@ -10,6 +10,11 @@ import {
     renderDropdown,
 } from "./common";
 
+function isSelectableLevelsDefined(selectableLevels) {
+    if (!selectableLevels || selectableLevels.length === 0) return undefined;
+    return selectableLevels;
```
tokland: [subjective] For very short logic expressions like this, I'd use a ternary instead of a guard (which are more used for longer functions that return early when preconditions are not met).

tokland (approved):
[code-only review]

PR #241, opened 2024-08-14
feature: new callback `onChildrenLoaded` to get children when new OrgUnit is selected

delemaf:
### :pushpin: References

-   **Issue:** [Closes](https://app.clickup.com/t/86957kfzj) #86957kfzj

### :memo: Implementation

New callback `onChildrenLoaded` to get children when new OrgUnit is selected `OrgUnitsSelector` component with the `selectableLevels` prop to limit the org units. Purpose of this was to get only displayed OrgUnits ids for filtering with selectableIds.

```tsx
const onChildrenLoaded = (orgUnits: OrgUnits[]) => useCallback({
	// Here you can filter with orgUnits props
	// Ex: filter on geometry
	setSelectableIds(orgUnits.filter(child => !!child.geometry).map(child => child.id))
}, [orgUnitsWithCoordinates])

<OrgUnitsSelector
    onChildrenLoaded={onChildrenLoaded}
	selectableIds={selectableIds}
 />
```

https://github.com/user-attachments/assets/3055ce5f-7064-44ea-abb8-4aa140561cab

### DIFF
--- a/src/org-unit-tree/OrgUnitTree.component.js
+++ b/src/org-unit-tree/OrgUnitTree.component.js
@@ -69,7 +69,7 @@ class OrgUnitTree extends React.Component {
     }
 
     setChildState(children) {
-        if (this.props.onChildrenLoaded) this.props.onChildrenLoaded(children);
+        if (this.props.onChildrenLoaded) this.props.onChildrenLoaded.fn(children);
 
         const keyToOrder = this.props.useShortNames ? "shortName" : "displayName";
 
@@ -80,7 +80,7 @@ class OrgUnitTree extends React.Component {
     }
 
     loadChildren() {
-        const { root, api, idsThatShouldBeReloaded } = this.props;
+        const { root, api, idsThatShouldBeReloaded, onChildrenLoaded } = this.props;
 
         if (
             (this.state.children === undefined && !this.state.loading) ||
@@ -89,19 +89,26 @@ class OrgUnitTree extends React.Component {
             this.setState({ loading: true });
 
             const childrenIds = root.children.map(({ id }) => id);
+            const fields = {
+                id: true,
+                level: true,
+                displayName: true,
+                shortName: true,
+                children: true,
+                path: true,
+                parent: true,
+            };
+
+            if (onChildrenLoaded?.fields) {
+                onChildrenLoaded.fields.forEach(field => {
+                    fields[field] = true;
+                });
+            }
+
             api.models.organisationUnits
                 .get({
                     paging: false,
-                    fields: {
-                        id: true,
-                        level: true,
-                        displayName: true,
-                        shortName: true,
-                        children: true,
-                        path: true,
-                        parent: true,
-                        geometry: true,
-                    },
+                    fields,
                     filter: {
                         id: {
                             in: childrenIds,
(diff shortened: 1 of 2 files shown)


ifoche:
Task linked: [CU-86957kfzj Filter out OU before asking for all points in "import rule"](https://app.clickup.com/t/86957kfzj)

src/org-unit-tree/OrgUnitTree.component.js
```diff
@@ -100,6 +100,7 @@ class OrgUnitTree extends React.Component {
                         children: true,
                         path: true,
                         parent: true,
+                        geometry: true,
```
tokland: [To evaluate with your PM]

This has a side-effect for everyone using the component: geometry will be requested unconditionally. With MultiPolygon geometries, that may increase responses sizes. Two approaches:

- [simple] Ask for field "geometry" only if `onChildrenLoaded` is requested (as the component itself does not need it)
- [more complex] Change the prop so it works like this: `onChildrenLoaded={ {fields: ["geometry"], fn: someFunction }}`

src/org-units-selector/OrgUnitsSelector.jsx
```diff
@@ -225,6 +227,10 @@ export default class OrgUnitsSelector extends React.Component {
         this.setState(state => ({
             roots: state.roots.map(r => (r.path === root.path ? mergeChildren(r, children) : r)),
         }));
+
+        if (this.props.onChildrenLoaded) {
+            this.props.onChildrenLoaded(children);
```
tokland: The root org unit is not notified, right? I guess for a first approach (and for your use case) this is not a problem.

delemaf: Yes only childrens 👍

src/org-unit-tree/OrgUnitTree.component.js
delemaf: I used the complex solution, it could be useful for optimisations of other fields 🚀

delemaf: image
I made sure it wasn't used in our code in case it break something

src/org-unit-tree/OrgUnitTree.component.js
```diff
@@ -89,19 +89,26 @@ class OrgUnitTree extends React.Component {
...
+                path: true,
+                parent: true,
+            };
+
+            if (onChildrenLoaded?.fields) {
+                onChildrenLoaded.fields.forEach(field => {
+                    fields[field] = true;
+                });
+            }
+
```
tokland: fields is an internal variable, so no big a big deal, but to be consistent with the FP/immutable style, we could refactor it with lodash. Proposal (untested): 

```typescript
            const extraFields = _(onChildrenLoaded.fields || [])
                .map(field => [field, true])
                .fromPairs()
                .value();

            const fields = {
                id: true,
                level: true,
                displayName: true,
                shortName: true,
                children: true,
                path: true,
                parent: true,
                ...extraFields,
            };
```

tokland (approved):
All good! :+1:

MiquelAdell (approved):

PR #243, opened 2024-12-02
Added slot for children in orgunit label and disabled state prop

p3rcypj:
### :pushpin: References

CLPC Facial reconstruction https://github.com/EyeSeeTea/emergency-responses-app/pull/70

### :memo: Implementation

- Added children slot for labels, in order to allow jsx next to the label
- Added disabled prop to disallow selecting an orgunit (for loading states where the component is used)

### :art: Screenshots

image
_Added Chip tag as an example_

image
_Disabled state; it still allow expanding the tree, but not selecting any orgunit_

### :fire: Testing

`docker.eyeseetea.com/samaritans/dhis2-data:40.4.1-sp-ip-develop-est`

In EMR app, go to `http://localhost:8081/#/frr` and select 'Responses Configuration'. If you don't see anything or it says you are not in the user group, you should add your user to the needed user groups. If you don't have user, ask to project manager.

image

On 'Responses configuration' you will see the orgunit tree. If you select a country and later click on 'Create response for country level' and confirm the dialog, you will see the disabled state while the request outside is loading. Check video media on the attached PR if needed.

### DIFF
--- a/src/org-unit-tree/OrgUnitTree.component.js
+++ b/src/org-unit-tree/OrgUnitTree.component.js
@@ -173,6 +173,7 @@ class OrgUnitTree extends React.Component {
                     currentRoot={this.props.currentRoot}
                     onChangeCurrentRoot={this.props.onChangeCurrentRoot}
                     labelStyle={this.props.labelStyle}
+                    labelChildren={this.props.labelChildren}
                     selectedLabelStyle={this.props.selectedLabelStyle}
                     arrowSymbol={this.props.arrowSymbol}
                     idsThatShouldBeReloaded={this.props.idsThatShouldBeReloaded}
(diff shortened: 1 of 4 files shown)


src/org-unit-tree/OrgUnitTree.component.js
```diff
@@ -90,21 +90,27 @@ class OrgUnitTree extends React.Component {
...
+                    id: true,
+                    level: true,
+                    displayName: true,
+                    shortName: true,
+                    children: true,
+                    path: true,
+                    parent: true,
+                },
+                extraFields
+            );
```
tokland, line 113: No need to change it, but typically: `{id: true, ...., ..extraFields}`

tokland (approved):
code-wise, all good

src/org-unit-tree/OrgUnitTree.component.js
p3rcypj: But `extraFields` might be `undefined`. Thats why I used `Object.assign`, I tried in console to do `{ hello:"a", ...undefined }` at console and it gave me error

tokland: In the browser console, object literals appear to require parentheses: `({ hello:"a", ...undefined })`. 

However, even if "undefined" were problematic, you'd typically use something like ' ...(extraFields || {})`.

PR #249, opened 2025-04-02
remove rxjs

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8698ekxj3

### :memo: Implementation

rxjs was around 800KB in our applications and it was being used only in 2 components: `<Sharing />` and `<MultiSelector />`

- In the Sharing component we use a debounce strategy to avoid making requests on every keystroke. I've replaced the rxjs debounce implementation with lodash.

- For the `<MultiSelector />` component it was being used to handle state so I've replaced with the `useState` hook.

- rxjs is a dependency in `dhis2/d2-ui-core` so I had to modify the way we were importing components from this library:
Doing this `import { TreeView } from "@dhis2/d2-ui-core";` was still including rxjs in our bundle. 
I've changed to `import TreeView from "@dhis2/d2-ui-core/tree-view/TreeView.component";` 

d2-ui-component current size in a new skeleton-app: **138 KB**
rxjs: **845 KB**

bundle_d2_ui_rxjs_detail

bundle_d2_ui_rxjs

d2-ui-component size after changes: **128.5 KB**

bundle_d2_ui_no_rxjs_detail

No more rxjs

bundle_d2_ui_no_rxjs

Javascript current size in skeleton app after build: **2.5 MB**

bundle_d2_ui_assets_rxjs

Javascript size in skeleton app after build with these changes: **1.9 MB**

bundle_d2_ui_assets_no_rxjs

[skeapp_d2_ui_components_norxjs.webm](https://github.com/user-attachments/assets/64777c83-bf84-413e-a14f-4636363d46bb)

You can test the skeleton app's build size in this branch. https://github.com/EyeSeeTea/dhis2-app-skeleton/tree/test-d2-ui-norxjs

You can use the play demo server to test the Sharing component: https://play.dhis2.org/41
username: admin
password: district

#8698ekxj3


ifoche:
Task linked: [CU-8698ekxj3 Feedback from 24/03/2025](https://app.clickup.com/t/8698ekxj3)

src/multi-selector/MultiSelector.tsx
```diff
@@ -0,0 +1,272 @@
+import React from "react";
+import {
+    Button,
+    IconButton,
+    Paper,
+    TextField,
+    makeStyles,
+    deprecatedPropType,
```
gqcorneby: I think this dependency is unused. We can remove it.

src/multi-selector/MultiSelector.tsx
```diff
@@ -0,0 +1,272 @@
...
+        moveItemsToLeft,
+        onSelectionChange,
+        updateItemsOnDoubleClick,
+        moveAllItems,
+        reorderSelectedItems,
+    } = useMultiSelectorMethods(props);
+
+    const placeholderText = searchFilterLabel
+        ? searchFilterLabel
+        : "Search available/selected items";
```
gqcorneby: Use i18n.t for texts. i18n.t(""Search available/selected items")

src/multi-selector/MultiSelector.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+    React.useEffect(() => {
+        onChange(optionItems.rightItems.map(item => item.value));
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [optionItems]);
```
gqcorneby: Use optionItems.rightItems as a dependency instead to avoid unintended onChange calls triggered by changes in rightSelected and leftSelected.

gqcorneby (changes requested):
Thank you, @eperedo! 

Updates in the orgtree and usersearch looks good! I tested MultiSelect in Bulk-Load (updated `"@eyeseetea/d2-ui-components": "../d2-ui-components/build/"` in package.json and force install). It seems like `optionItems` is not getting updated properly when either props `options` or `selected` change. In the video, I'm printing options and selected length.
1. after selecting `Antenatal care visit`, all 3 should be selected but the selector shows none selected
2. by the end of the video additional items were being displayed

https://github.com/user-attachments/assets/d1db34e8-bb1a-4737-b9a4-73de5603d8c1

src/multi-selector/SelectMultiple.tsx
```diff
@@ -0,0 +1,75 @@
...
+        onSelectionChange(values);
+    };
+
+    const handleDoubleClick = (e: React.MouseEvent<HTMLOptionElement>) => {
+        onDoubleClick(e.currentTarget.value);
+    };
+
+    const orderedItems = ordered
+        ? _(items)
+              .sortBy(x => x.text)
```
gqcorneby: I'm not sure if we want to impose this sorting. Seems like the previous version uses the options input as is. One use case use of not wanting this sort is if the options are months for example, we wouldn't want to sort that by text.

src/multi-selector/hooks.tsx
```diff
@@ -0,0 +1,233 @@
...
+            rightItems: uniqueOptions.filter(option => selected.includes(option.value)),
+            rightSelected: [],
+            leftSelected: [],
+        };
+    });
+
+    React.useEffect(() => {
+        setOptionItems(prev => ({
+            ...prev,
+            rightItems: uniqueOptions.filter(option => selected.includes(option.value)),
```
gqcorneby: This is kind of related to my previous comment. Currently, `leftItems` is displayed sorted by text, while we're using uniqueOptions as the reference to build the right items, causing a mismatch in sorting between the two sides. 

This will also rebuild the `rightItems` when `onChange` is called, which results in the right items being sorted back to the original order (from uniqueOptions) rather than the input order. This also interferes with the behavior of `reorderSelectedItems`, as it reverts the list to whatever order `uniqueOptions` defines.

I think we just need to sort it by selected index
```
_.sortBy(
      _.filter(uniqueOptions, option => selected.includes(option.value)),
      option => _.indexOf(selected, option.value)
  )
```

gqcorneby (changes requested):
One last thing @eperedo! Just some sorting issues and I think we're good.

gqcorneby (approved):
Working great! Thank you @eperedo

PR #251, opened 2025-09-22
add disabled property to TableColumn 

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869a4vv2h

### :memo: Implementation

- [x] add disabled property avoid columns being removed
- [x] add children prop to allow custom components after Transfer component

User columns selector for User-Extended

image

can you please help me with a beta version?

#869a4vv2h

### DIFF
--- a/src/data-table/ColumnSelectorDialog.tsx
+++ b/src/data-table/ColumnSelectorDialog.tsx
@@ -8,13 +8,17 @@ import { TableColumnSelector } from "./TableColumnSelector";
 
 interface ColumnSelectorDialogProps<T extends ReferenceObject> {
     columns: TableColumn<T>[];
-    visibleColumns: (keyof T)[];
+    visibleColumns: TableColumnsType<T>;
     allowReorderingColumns?: boolean;
-    onChange: (visibleColumns: (keyof T)[]) => void;
+    onChange: (visibleColumns: TableColumnsType<T>) => void;
     onCancel: () => void;
     childrenTransfer?: React.ReactNode;
     keepDisabledColumns?: boolean;
 }
+
+type TableColumnsType<T> = (keyof T)[];
+
+type UpdateSelectedColumns<T> = (params: { selected: TableColumnsType<T> }) => void;
 
 export function ColumnSelectorDialog<T extends ReferenceObject>(
     props: ColumnSelectorDialogProps<T>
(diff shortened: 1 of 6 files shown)


src/data-table/ColumnSelectorDialog.tsx
```diff
@@ -11,13 +11,25 @@ interface ColumnSelectorDialogProps<T extends ReferenceObject> {
...
-    const sortableColumns = columns.map(({ name, text: label }) => ({ label, value: name }));
+    const {
+        childrenTransfer,
+        columns,
+        visibleColumns,
+        onChange,
+        onCancel,
+        allowReorderingColumns = true,
+    } = props;
+    const sortableColumns = columns.map(({ name, text: label, disabled }) => ({
```
tokland: Now that we are changing this, let's specify the return type of the block (`TransferOption`) so we can validate it and navigate to the documentation directly.

tokland (approved):
Beta released: v2.12.0-beta.1

I just added a very minor proposal (type-only), so PR approved. No need to ask for re-review unless more changes are added.

eperedo:
Requesting a review again because I've added a prop to keep disabled columns selected.

https://github.com/user-attachments/assets/1c48542a-c532-483b-9183-62f93d305551

src/data-table/ColumnSelectorDialog.tsx
```diff
@@ -24,12 +26,33 @@ export function ColumnSelectorDialog<T extends ReferenceObject>(
...
+        })
+    );
+
+    const selectedColumns = React.useMemo((): string[] => {
+        const disableColumns = columns.filter(col => col.disabled).map(col => col.name.toString());
+        return _(visibleColumns).map(String).concat(disableColumns).uniq().value();
+    }, [visibleColumns, columns]);
+
+    const updateSelectedColumns = React.useCallback(
+        ({ selected }: { selected: Array<keyof T> }) => {
```
tokland: Nice to have this immutable.

[minor] It's probably more idiomatic to write `useCallback<T>`, being T a function of the prop type (using NonNullable, if it's optional).

src/data-table/ColumnSelectorDialog.tsx
```diff
@@ -24,12 +26,33 @@ export function ColumnSelectorDialog<T extends ReferenceObject>(
...
+    const sortableColumns = columns.map(
+        ({ name, text: label, disabled }): TransferOption => ({
+            label,
+            value: name.toString(),
+            disabled: disabled ?? false,
+        })
+    );
+
+    const selectedColumns = React.useMemo((): string[] => {
+        const disableColumns = columns.filter(col => col.disabled).map(col => col.name.toString());
```
tokland: In the useCallback we check keepDisabledColumns to build the final value. We can apply the same logic in the useMemo, allowing selectedColumns to be passed directly to selected.

tokland (approved):
Beta: v2.12.0-beta.2

Minor things:

src/data-table/ColumnSelectorDialog.tsx
```diff
@@ -8,14 +8,18 @@ import { TableColumnSelector } from "./TableColumnSelector";
...
+    visibleColumns: TableColumnsType<T>;
     allowReorderingColumns?: boolean;
-    onChange: (visibleColumns: (keyof T)[]) => void;
+    onChange: (visibleColumns: TableColumnsType<T>) => void;
     onCancel: () => void;
     childrenTransfer?: React.ReactNode;
     keepDisabledColumns?: boolean;
 }
 
+type TableColumnsType<T> = (keyof T)[];
```
tokland, line 19: [no need to change, just a note] When one is forced to use parens, Array<...> is more idiomatic.

tokland (approved):
Beta release: v2.12.0-beta.3

PR #252, opened 2025-10-11
Feature/object table custom config

gqcorneby:
### :pushpin: References

-   **Issue:** Closes #869a4vv2h [Row and column configuration is missing](https://app.clickup.com/t/869atk4wy)

### :memo: Implementation
- add optional parameter to render custom config in object table menu

Additional:
- add disabled property in dropdown options (I copied the component in dataset-configuration to add the option to disable. I included the update here since it might be useful for others)

**custom config component rendered in table actions**
image

**disabled dropdown options**
image

### DIFF
--- a/src/data-table/DataTable.tsx
+++ b/src/data-table/DataTable.tsx
@@ -107,7 +107,7 @@ export interface DataTableProps<T extends ReferenceObject> {
     customClasses?: CustomClassesType;
     stickyHeader?: boolean;
     columnsAlignment?: "left" | "center" | "right";
-    customConfig?: React.ReactNode;
+    globalActionComponents?: React.ReactNode;
     childrenTransfer?: React.ReactNode;
     keepDisabledColumns?: boolean;
 }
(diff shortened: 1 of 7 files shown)


src/data-table/DataTable.tsx
```diff
@@ -107,6 +107,7 @@ export interface DataTableProps<T extends ReferenceObject> {
     customClasses?: CustomClassesType;
     stickyHeader?: boolean;
     columnsAlignment?: "left" | "center" | "right";
+    customConfig?: React.ReactNode;
```
eperedo: Since this prop is exclusive to rendering a react component for global actions I think we should use a different name instead `customConfig`. We have a similar prop `filterComponents`, so maybe this could be `globalActionComponents`?

eperedo (changes requested):

eperedo (approved):
There's a conflict to solve in the package.json, but code looks good to me. Thanks @gqcorneby!

MiquelAdell:
the conflict was easy to solve so I did and merging and approving

PR #255, opened 2025-11-14
fix/Dropdown use generic types

gqcorneby:
### :pushpin: References

-   **Issue:** Closes: #869b2wa0d https://app.clickup.com/t/869b2wa0d

### :memo: Implementation
- Updated dropdown component to accept generic type. Props already support generic type for options and value.

### DIFF
--- a/src/dropdown/Dropdown.tsx
+++ b/src/dropdown/Dropdown.tsx
@@ -28,7 +28,7 @@ const SelectWrapper: React.FC<SelectProps> = React.memo(props => {
     );
 });
 
-export const DropdownInner = <Value extends string = string>(props: DropdownProps<Value>) => {
+const DropdownInner = <Value extends string = string>(props: DropdownProps<Value>) => {
     const { items, value, onChange, label, hideEmpty, className } = props;
 
     const selectValue =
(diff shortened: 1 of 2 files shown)


src/dropdown/Dropdown.tsx
```diff
@@ -5,7 +5,7 @@ import { DropdownForm, DropdownItem } from "./GenericDropdown";
 
 export interface DropdownProps<Value extends string = string> {
     className?: string;
-    items: DropdownItem[];
+    items: DropdownItem<Value>[];
```
tokland, line 8: Cool, definitely the generic type should apply also to the items.

src/dropdown/Dropdown.tsx
```diff
@@ -54,4 +54,8 @@ export const Dropdown: React.FC<DropdownProps> = React.memo(props => {
             </Select>
         </SelectWrapper>
     );
-});
+};
+
+export const Dropdown = React.memo(DropdownInner) as <Value extends string = string>(
```
tokland: Yep, it's a known problem, React.memo drops the generic type. The typical workaround, DRYer: 

```export const Dropdown = React.memo(DropdownInner) as typeof DropdownInner;```

tokland (approved):

PR #256, opened 2025-11-23
only include org. units within the users scope in filter by level/group

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869b74ga4

### :memo: Implementation

added a new prop `withinUserHierarchyInFilters` to allow selection by level/group to only include org. units within the users scope (right now is including every org. unit in the selected level/group)

https://github.com/user-attachments/assets/bf26d7af-6a81-4399-8b63-50f62efc1f4d

can you please help me with a beta version so I can create the PR in dataset-configuration? @MiquelAdell 

#869b74ga4

### DIFF
--- a/src/org-unit-select/OrgUnitSelectByGroup.component.js
+++ b/src/org-unit-select/OrgUnitSelectByGroup.component.js
@@ -126,17 +126,24 @@ class OrgUnitSelectByGroup extends React.Component {
     componentDidMount() {
         const { api } = this.context;
         if (this.props.withinUserHierarchyInFilters) {
-            api.get("/organisationUnits/", {
-                paging: false,
-                fields: "id",
-                withinUserHierarchy: this.props.withinUserHierarchyInFilters,
-            })
+            this.setState({ loading: true });
+            api.models.organisationUnits
+                .get({
+                    paging: false,
+                    fields: { id: true },
+                    withinUserHierarchy: this.props.withinUserHierarchyInFilters,
+                })
                 .getData()
                 .then(response => {
-                    if (response.organisationUnits.length > 0) {
-                        const orgUnitIdsSet = new Set(response.organisationUnits.map(ou => ou.id));
+                    if (response.objects.length > 0) {
+                        const orgUnitIdsSet = new Set(response.objects.map(ou => ou.id));
                         this.setState({ orgUnitsInHierarchy: orgUnitIdsSet });
                     }
+                    this.setState({ loading: false });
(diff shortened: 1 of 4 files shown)


MiquelAdell (approved):

src/org-unit-select/OrgUnitSelectByGroup.component.js
```diff
@@ -113,6 +122,24 @@ class OrgUnitSelectByGroup extends React.Component {
         // Material-UI SelectField will change height depending on whether or not it has a value
         return renderDropdown.call(this, menuItems, label);
     }
+
+    componentDidMount() {
+        const { api } = this.context;
+        if (this.props.withinUserHierarchyInFilters) {
```
gqcorneby, line 128: [not tested] do we also want to add a loading state here to avoid possible timing issues when `getOrgUnitsForGroup` si called?

src/org-unit-select/OrgUnitSelectByGroup.component.js
```diff
@@ -113,6 +122,24 @@ class OrgUnitSelectByGroup extends React.Component {
...
+            api.get("/organisationUnits/", {
```
gqcorneby: can we use `api.models.organisationUnits` here?

gqcorneby (commented):
Thanks @eperedo! Just some minor things.

I also contemplated if we needed to filter based on `orgUnitsInHierarchy` in `OrgUnitSelectByGroup.getOrgUnitsForGroup` in the branch ` if (this.props.currentRoot)`, but I think it shouldn't be necessary since `currentRoot` would be accessible by the user, right? And we wouldn't have orgUnits under this level that wouldn't be accessible?

eperedo:
> Thanks @eperedo! Just some minor things.
> 
> I also contemplated if we needed to filter based on `orgUnitsInHierarchy` in `OrgUnitSelectByGroup.getOrgUnitsForGroup` in the branch ` if (this.props.currentRoot)`, but I think it shouldn't be necessary since `currentRoot` would be accessible by the user, right? And we wouldn't have orgUnits under this level that wouldn't be accessible?

Yes, the user already have access to currentRoot, but also we cannot apply the param `orgUnitsInHierarchy` in the `/organisationUnits/{id}` endpoint. This filter only works for the `/organisationUnits` endpoint.

gqcorneby (approved):
All good! Thanks @eperedo!

PR #258, opened 2026-01-28
apply withinUserHierarchy param to search

MatiasArriola:
Needed for: https://app.clickup.com/t/869b5yvxv #869b5yvxv

Updates the `OrgUnitSelector`. If the prop `withinUserHierarchyInFilters` is true, then the `withinUserHierarchy` param will be used to query org units during search. 

`withinUserHierarchyInFilters` was being used for level/group filters, and this change also adds it for search.


Ramon-Jimenez (approved):
Works OK! Thanks @MatiasArriola

PR #260, opened 2026-05-11
[feature]: Display selection count notifications in ObjectsTable

deeonwuli:
### :pushpin: References

-   **Issue:** Closes:  https://app.clickup.com/t/869cqxtfn

### :memo: Implementation
 The d2-ui-components selection banner is only triggered when items are selected on *other* pages, so a user selecting rows on the current page gets no visual feedback before applying a bulk action.

  - Added `useTableWithSelectionCount` hook that wraps `useObjectsTable`
  - When items are selected only on the current page (and the existing banner would not appear), a notification is shown: *"There are {{count}} items selected on this page."* with a *Clear selection* link
  - The hook suppresses this notification when the standard DataTable selection messages already apply (cross-page selection, all items selected, etc.)

Propagated from https://github.com/EyeSeeTea/user-extended-app/pull/344
  
#869bryznd

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2025-04-02T23:36:26.113Z\n"
-"PO-Revision-Date: 2025-04-02T23:36:26.113Z\n"
+"POT-Creation-Date: 2026-05-11T15:35:33.204Z\n"
+"PO-Revision-Date: 2026-05-11T15:35:33.204Z\n"
 
 msgid "Cancel"
 msgstr ""
@@ -108,6 +108,11 @@ msgstr ""
 
 msgid "Search by name"
 msgstr ""
+
+msgid "There are {{count}} items selected on this page."
+msgid_plural "There are {{count}} items selected on this page."
+msgstr[0] ""
+msgstr[1] ""
 
 msgid "Select all"
 msgstr ""
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -110,9 +110,10 @@ msgid "Search by name"
 msgstr ""
 
 msgid "There is 1 item selected on this page."
-msgid_plural "There are {{count}} items selected on this page."
-msgstr[0] ""
-msgstr[1] ""
+msgstr ""
+
+msgid "There are {{count}} items selected on this page."
+msgstr ""
 
 msgid "Select all"
 msgstr ""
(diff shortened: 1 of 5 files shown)


i18n/es.po
```diff
@@ -112,6 +112,12 @@ msgstr ""
 msgid "Search by name"
 msgstr "Buscar por nombre"
 
+#, fuzzy
+msgid "There are {{count}} items selected on this page."
+msgid_plural "There are {{count}} items selected on this page."
+msgstr[0] "Hay {{total}} elementos seleccionados en todas las páginas."
```
eperedo: I think the variable should be {{count}} instead {{total}}, right? 

(same for fr.po)

i18n/en.pot
```diff
@@ -109,6 +109,11 @@ msgstr ""
 msgid "Search by name"
 msgstr ""
 
+msgid "There are {{count}} items selected on this page."
+msgid_plural "There are {{count}} items selected on this page."
+msgstr[0] ""
```
eperedo: Should we include specific translations for singular/plural cases?

"There is 1 item selected on this page"
"There are 2 items selected on this page"

(same for other languages in case you are agree with this)

src/objects-list/objects-list-hooks.ts
```diff
@@ -134,3 +136,52 @@ export function useObjectsTable<Obj extends ReferenceObject>(
...
+
+        const rows = tableProps.rows;
+        const total = tableProps.pagination?.total ?? rows.length;
+        const ids = tableProps.ids ?? [];
+
+        const selectionInOtherPages = _.differenceBy(selection, rows, "id");
+        const allSelectedInPage =
+            rows.length > 0 && _.differenceBy(rows, selection, "id").length === 0;
+        const multiplePagesAvailable = total > rows.length;
+        const selectAllImplemented = ids.length > 0;
```
eperedo: all of these variables: selectionInOtherPages, allSelectedInPage, multiplePagesAvailable, selectAllImplemented

follow the same logic that we already have in `selection.tsx` in the method `getSelectionMessages`.

Maybe we can refactor to avoid duplication?

eperedo (changes requested):

MiquelAdell:
@deeonwuli there are comments to address in this PR.
When you do, can you check if we need to change something here?
https://github.com/EyeSeeTea/user-extended/pull/344

src/objects-list/objects-list-hooks.ts
```diff
@@ -134,3 +137,44 @@ export function useObjectsTable<Obj extends ReferenceObject>(
...
+        const total = tableProps.pagination?.total ?? rows.length;
+        const ids = tableProps.ids ?? [];
+
+        const hasSelectionNotifications =
+            getSelectionMessages(rows, selection, total, ids, [], undefined).length > 0;
+        if (hasSelectionNotifications) return [];
+
+        return [
+            {
+                message: i18n.t("There is 1 item selected on this page.", {
```
eperedo: We're ignoring the count variable and using 1 for every selected item.

eperedo (changes requested):

eperedo (approved):

